PR #76, opened 2015-01-27
rbd volumes diff transferring in ceph-ceph case

toha10:
Import - Export full data between ceph clusters was changed to diff import - export


Wasapl:
Merged build finished. Test PASSed.  http://jenkins-cf.vm.mirantis.net:8080/job/CloudFerry_PR_checks/55/

PR #95, opened 2015-02-25
implement SAIC-40 : Add logging to scheduler in CF

oleksii-shyman:
- all logging configs moved to configs/logging_config.yml
  - added logging exception in scheduler
  - with call migrate:debug=True logger shows debug messages
  - added stream output with error log level to error.log
  - added stream output with info level to info.log


Wasapl:
Merged build finished. Test PASSed.  http://jenkins-cf.vm.mirantis.net:8080/job/CloudFerry_PR_checks/75/

PR #106, opened 2015-03-04
Implement SAIC-23 : add rollbacks to CF scheduler

oleksii-shyman:
- added multichain scheduler support
  - extended scenario file with additional chains

### DIFF
--- a/cloud/os2os.py
+++ b/cloud/os2os.py
@@ -102,14 +102,21 @@ class OS2OSFerry(cloud_ferry.CloudFerry):
                 utl.INSTANCES_TYPE: {}
             }
         })
+        # "process_migration" is dict with 3 keys:
+        #    "preparation" - is cursor that points to tasks must be processed
+        #                    before migration i.e - taking snapshots,
+        #                    figuring out all services are up
+        #    "migration" - is cursor that points to the first
+        #                  task in migration process
+        #    "rollback" - is cursor that points to tasks must be processed
+        #                 in case of "migration" failure
         if not scenario:
-            process_migration = self.process_migrate()
+            process_migration = {"migration": cursor.Cursor(self.process_migrate())}
         else:
             scenario.init_tasks(self.init)
             scenario.load_scenario()
-            process_migration = scenario.get_net()
-        process_migration = cursor.Cursor(process_migration)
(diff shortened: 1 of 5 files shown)


Wasapl:
Merged build finished. Test PASSed.  http://jenkins-cf.vm.mirantis.net:8080/job/CloudFerry_PR_checks/91/

roman-verchikov:
Global comment: why don't we move the functionality introduced here into

roman-verchikov:
Why don't we move the functionality introduced in `BaseScheduler` to a derived class (as suggested by Dmitriy Sozinov)? The changes here touch the core of CloudFerry, we should minimize the impact on other CloudFerry users.

mirrorcoder:
Code is ok,  i like. What specific functionality you want to move and at what place in BaseScheduler ?

roman-verchikov:
> mirrorcoder commented 8 days ago
> Code is ok, i like. What specific functionality you want to move and at what place in BaseScheduler ?

@mirrorcoder, please disregard my previous comment. I was trying to find a way if we can move the introduced functionality into a separate class rather than modifying existing. But after a conversation with @oleksii-shyman we decided there is no good way of doing this, so +1 from my side.

Settis:
Why we rolled back only destination cloud? CF also changes source cloud. We must take it info account.

roman-verchikov:
@Settis, why do you think this change will not rollback the source cloud? The changes made should work for any scenario with "ROLLBACK" and "PREPARATION" chains, that's basically why the changes are in the low-level scheduler base class.

Settis:
Let's take a look at [migrate.yaml and tasks.yaml](https://github.com/MirantisWorkloadMobility/CloudFerry/pull/106/files#diff-cd9e621f0ffd4f41fa1668781f3ab432R5). We do "create_snapshot" in "preparation" section and inicialize  it only for destination cloud (you can see it in tasks.yaml). And nothing else for source cloud.

Settis:
We discussed with Roman. Source cloud rollback will be done in separate issue.
+1 from my side.

roman-verchikov:
@Settis: 
1. We'll work on source cloud rollback later (there's a task for that already)
2. We also have a task which covers VM rollback/deletion on destination
3. We expect that destination objects (e.g. neutron ports) will be deleted by ovs-agent based on new DB data

With that in mind let's merge this patch unless you see more issues in the patch which require our attention.

PR #117, opened 2015-03-13
Fix bug with the 'time_wait' parameter

miarmak:
Since 'time_wait' is defined as IntOpt in the cfglib.py, it should have
integer value, instead of string value. In other case it raises
exception, when a Proxy wrapper is used for OpenStack clients in the
CloudFerry resources.


Wasapl:
Merged build finished. Test PASSed.  http://jenkins-cf.vm.mirantis.net:8080/job/CloudFerry_PR_checks/111/

mirrorcoder:
+1

oleksii-shyman:
looks ok

roman-verchikov:
+1

PR #122, opened 2015-03-17
Private flavors

toha10:
Added checking for type of flavor (public/private). In case of private flavors added working with permissions (allowed tenants for flavor).

### DIFF
--- a/cloudferrylib/os/compute/nova_compute.py
+++ b/cloudferrylib/os/compute/nova_compute.py
@@ -89,8 +89,8 @@ class NovaCompute(compute.Compute):
         for keypair in self.get_keypair_list():
             info['keypairs'][keypair.id] = self.convert(keypair)
 
-        for flavor in self.get_flavor_list():
-            info['flavors'][flavor.id] = self.convert(flavor)
+        for flavor in self.get_flavor_list(is_public=None):
+            info['flavors'][flavor.id] = self.convert(flavor, cloud=self.cloud)
 
         if self.config.migrate.migrate_quotas:
             self._read_info_quotas(info)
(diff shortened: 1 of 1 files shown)


Wasapl:
Merged build finished. Test PASSed.  http://jenkins-cf.vm.mirantis.net:8080/job/CloudFerry_PR_checks/119/

miarmak:
Looks ok.

cloudferrylib/os/compute/nova_compute.py
```diff
@@ -89,8 +89,8 @@ def _read_info_resources(self, **kwargs):
         for keypair in self.get_keypair_list():
             info['keypairs'][keypair.id] = self.convert(keypair)
 
-        for flavor in self.get_flavor_list():
-            info['flavors'][flavor.id] = self.convert(flavor)
+        for flavor in self.get_flavor_list(is_public=None):
```
roman-verchikov: 1. This looks counter-intuitive. `is_public` looks like boolean variable, while for some reason you assign it to `None` rather than `True` or `False`
2. `get_flavor_list` is always used with `is_public=None` throughout the code, why not setting it directly in the `get_flavor_list()`? In other words instead of modifying all the calls to `get_flavor_list`, just modify the internals of the method.

roman-verchikov:
Please fix inline

toha10:
1. If we want to get all flavors (including private), we need to set is_public=None
from novaclient source code (https://github.com/openstack/python-novaclient/blob/2.15.0/novaclient/v1_1/flavors.py#L92)
# is_public is ternary - None means give all flavors.

2. I think that it can be used as an option, in this case we need to leave it as argument.
We can have a case that we need to migrate only public flavors w/o some specific private flavors for some tenants as a some unwanted trash. But may be i'm wrong.
How do you think?

roman-verchikov:
> If we want to get all flavors (including private), we need to set is_public=None from novaclient source code (https://github.com/openstack/python-novaclient/blob/2.15.0/novaclient/v1_1/flavors.py#L92)
OK

> I think that it can be used as an option, in this case we need to leave it as argument.
> We can have a case that we need to migrate only public flavors w/o some specific private flavors for some tenants as a some unwanted trash. But may be i'm wrong.
> How do you think?
I think YAGNI :) Let's try *not* modifying method signature in first place

PR #136, opened 2015-03-27
Changes to config.ini in development branch

oleksii-shyman:
- changes according to vagrant devlab
  - added keystone/mysql settings
  - disabled swift usage in configs (depends on SAIC-66)


Wasapl:
Merged build finished. Test PASSed.  http://jenkins-cf.vm.mirantis.net:8080/job/CloudFerry_PR_checks/165/

miarmak:
I guess we need to create separate config for such purposes, because config.ini supposed to be smth like a config template without any specific information.

oleksii-shyman:
From one side - it is fine as soon as we are in devel branch
from other side - we will have problems in the future when we will merge changes to master branch (because we will have to exclude commit from history and it will break our fast-forward merge plan)

miarmak:
Why do we need exclude any commits now? We can make smth like configs/devlab_config.ini and merge it to master too.

PR #137, opened 2015-03-28
To avoid DHCP taking IP used by instance

wzbbbb:
Since the creation order of a DHCP port and an instance port can be
random, sometimes, the DHCP port can take the IP used by an instance
in source could. To avoid it, this fix will disable DHCP and free its
IP before instance port creation, then re-enable the DHCP.


Wasapl:
Merged build finished. Test FAILed.   http://jenkins-cf.vm.mirantis.net:8080/job/CloudFerry_PR_checks/167/

roman-verchikov:
Create pull request against `devel` branch, not `master`

PR #140, opened 2015-03-30
 SAIC-97, To avoid DHCP taking IP used by an instance

wzbbbb:
Since the creation order of a DHCP port and an instance port can be
random, sometimes, the DHCP port can take the IP used by an instance
in source could. To avoid it, this fix will disable DHCP and free its
IP before instance port creation, then re-enable the DHCP.

### DIFF
--- a/cloudferrylib/os/actions/prepare_networks.py
+++ b/cloudferrylib/os/actions/prepare_networks.py
@@ -31,6 +31,13 @@ class PrepareNetworks(action.Action):
         keep_ip = self.cfg.migrate.keep_ip
 
         instances = info_compute[utl.INSTANCES_TYPE]
+        # disable DHCP in all subnets
+        subnets = network_resource.get_subnets()
+        for snet in subnets:
+            if snet['tenant_name'] == self.cfg.dst.tenant:
+                network_resource. \
+                    reset_subnet_dhcp(snet['id'], False)
+
         for (id_inst, inst) in instances.iteritems():
             params = []
             networks_info = inst[utl.INSTANCE_BODY][utl.INTERFACES]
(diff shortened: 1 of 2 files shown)


Wasapl:
Merged build finished. Test PASSed.  http://jenkins-cf.vm.mirantis.net:8080/job/CloudFerry_PR_checks/177/

roman-verchikov:
+1

cloudferrylib/os/actions/prepare_networks.py
```diff
@@ -31,6 +31,13 @@ def run(self, info=None, **kwargs):
         keep_ip = self.cfg.migrate.keep_ip
 
         instances = info_compute[utl.INSTANCES_TYPE]
+        # disable DHCP in all subnets
+        subnets = network_resource.get_subnets()
+        for snet in subnets:
+            if snet['tenant_name'] == self.cfg.dst.tenant:
```
miarmak: What if we specify admin tenant and want to migrate all VMs?

cloudferrylib/os/actions/prepare_networks.py
```diff
@@ -31,6 +31,13 @@ def run(self, info=None, **kwargs):
...
+                network_resource. \
```
miarmak: Could you please do not use line breaks if it is possible?

miarmak:
1 question and 1 little style issue inline)

cloudferrylib/os/actions/prepare_networks.py
roman-verchikov: This is actually not required (at least for pep8 and flake8 with default configuration). Looks like this is only openstack style requirement, which we can skip.

miarmak: OK, but is not a reason, why I didn't wrote my "+1"))

cloudferrylib/os/actions/prepare_networks.py
wzbbbb: We will need to decide which keywords are allowed, then update it. Right now, it matches the way the config.ini works.
For keywords, we may also want to allow regex or a '*' in it.

cloudferrylib/os/actions/prepare_networks.py
wzbbbb: it is longer that 80.

miarmak:
Ok, l do not have objections) Thanks! +1

PR #145, opened 2015-04-02
Fix for __patch_option_bootable_of_volume method.

toha10:
Fix provides __patch_option_bootable_of_volume (cinder resource) method with mysqlconnector, because it used "not direct" interaction with DB (via ssh to DB node and performing bash command).
In some cases it could lead to failure, because we can't connect to DB host from mysql section of config via ssh:
f.e. in case of HA architecture, we can connect to DB through Public VIP with mysqlconnector, but we can't ssh to this ip address.

### DIFF
--- a/cloudferrylib/os/storage/cinder_storage.py
+++ b/cloudferrylib/os/storage/cinder_storage.py
@@ -15,9 +15,6 @@
 
 import time
 
-from fabric.api import run
-from fabric.api import settings
-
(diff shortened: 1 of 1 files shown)


roman-verchikov:
Less code is better code, +1

Wasapl:
Merged build finished. Test PASSed.  http://jenkins-cf.vm.mirantis.net:8080/job/CloudFerry_PR_checks/195/

PR #147, opened 2015-04-07
SAIC-95: Add flag 'all_vms' to config

miarmak:
Add availability to migrate all VMs with flag 'all_vms=True' in the
'migrate' section of config. It is possible if destination user,
specified in the 'dst' section of config, has admin role in all
destination tenants, otherwise it will fail with 'Unauthorized' error
(Actually, it was required by CloudFerry before to migrate instances
to the different from 'admin' tenant).

### DIFF
--- a/cfglib.py
+++ b/cfglib.py
@@ -107,6 +107,10 @@ migrate_opts = [
                help='Path to file with the groups of VMs'),
     cfg.BoolOpt('all_networks', default=False,
                 help="Migrate all network resources from all tenants"),
+    cfg.BoolOpt('all_vms', default=False,
+                help="Migrate all VM's from all tenants. User, specified in "
+                     "the 'dst' section of config also should have admin role "
+                     "in all tenants."),
 ]
 
 mail = cfg.OptGroup(name='mail',
--- a/cloudferrylib/os/compute/nova_compute.py
+++ b/cloudferrylib/os/compute/nova_compute.py
@@ -114,6 +114,11 @@ class NovaCompute(compute.Compute):
             raise ValueError('Only "resources" or "instances" values allowed')
 
         search_opts = kwargs.get('search_opts')
+
+        if self.config.migrate.all_vms:
+            search_opts = search_opts if search_opts else {}
+            search_opts.update(all_tenants=True)
+
         info = {'instances': {}}
 
         for instance in self.get_instances_list(search_opts=search_opts):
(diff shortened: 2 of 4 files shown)


Wasapl:
Merged build finished. Test PASSed.  http://jenkins-cf.vm.mirantis.net:8080/job/CloudFerry_PR_checks/200/

cfglib.py
```diff
@@ -105,6 +105,10 @@
                help='Size of one chunk to transfer via SSH'),
     cfg.StrOpt('group_file_path',
                help='Path to file with the groups of VMs'),
+    cfg.BoolOpt('all_vms', default=False,
+                help="Migrate all VM's from all tenants. User, specified in "
```
roman-verchikov: This says "Migrate all VMs from all tenants", do we ever check if user specified in config has admin privileges to migrate all VMs? Or will we migrate some subset of VMs user has access to, and then get error?

miarmak: No, we do not check it. But we have it in our requirements (https://github.com/MirantisWorkloadMobility/CloudFerry/blob/devel/README.md):
`Credentials of global cloud admin for both clouds.`
So we assume, that this user has admin privileges.

Wasapl:
Merged build finished. Test FAILed.   http://jenkins-cf.vm.mirantis.net:8080/job/cloudferry-unittests-and-style-checks/11/

Wasapl:
Merged build finished. Test PASSed.  http://jenkins-cf.vm.mirantis.net:8080/job/cloudferry-unittests-and-style-checks/12/

oleksii-shyman:
Looks ok + 1

Wasapl:
Merged build finished. Test PASSed.  http://jenkins-cf.vm.mirantis.net:8080/job/cloudferry-unittests-and-style-checks/27/

miarmak:
Rebased

roman-verchikov:
Looks ok, but I still have concerns regarding `all_vms`, `all_networks`, etc. config options. Wouldn't it be better to have a migration tenant specified in config, and then migration would simply migrate everything the specified tenant has access to?

miarmak:
It would be really nice to have a migration tenant specified in config (Moreover, we already have this task in the "Nice to have" epic, but as we decided, it is not in our current development scope). 

But now we do not have such possibility and we have issues, when we specify 'admin' tenant. So we actually can not define currently, rather we want to migrate all that we have with the 'admin' tenant credentials or we want to migrate stuff only related to this 'admin' tenant.

roman-verchikov:
+1

PR #158, opened 2015-04-15
SAIC-69: Fix bug with the Floating IPs quota excess

miarmak:
Different versions of quantum/neutron clients raise different types of
exceptions on the same error (floating ip quota exceed). Also, they use
absolutely different error messages. From now we will handle it with
more general exception and status code check.

### DIFF
--- a/cloudferrylib/os/network/neutron.py
+++ b/cloudferrylib/os/network/neutron.py
@@ -15,7 +15,7 @@
 
 import ipaddr
 
-from neutronclient.common.exceptions import IpAddressGenerationFailureClient
+from neutronclient.common import exceptions as neutron_exc
 from neutronclient.v2_0 import client as neutron_client
 
 from cloudferrylib.base import network
@@ -950,9 +950,12 @@ class NeutronNetwork(network.Network):
             while True:
                 self.neutron_client.create_floatingip({
                     'floatingip': {'floating_network_id': ext_net_id}})
-        except IpAddressGenerationFailureClient:
-            LOG.info("| Floating IPs "
-                     "were allocated in network %s" % ext_net_id)
+        except neutron_exc.NeutronClientException as e:
+            if e.status_code == 409:  # 409 - Conflict
+                LOG.info("| Floating IPs were allocated in network %s" %
+                         ext_net_id)
+            else:
+                raise
 
     def recreate_floatingips(self, src_floats, src_nets,
                              existing_nets,


Wasapl:
Merged build finished. Test PASSed.  http://jenkins-cf.vm.mirantis.net:8080/job/cloudferry-unittests-and-style-checks/18/

cloudferrylib/os/network/neutron.py
```diff
@@ -950,9 +950,12 @@ def allocate_floatingips(self, ext_net_id):
...
                     'floatingip': {'floating_network_id': ext_net_id}})
-        except IpAddressGenerationFailureClient:
-            LOG.info("| Floating IPs "
-                     "were allocated in network %s" % ext_net_id)
+        except neutron_exc.NeutronClientException as e:
+            if e.status_code == 409:  # 409 - Conflict
+                LOG.info("| Floating IPs were allocated in network %s" %
+                         ext_net_id)
+            else:
+                raise
```
oleksii-shyman: raise is bad practice
raise RuntimeError("Neutron ERROR not related to floating ips") works better

miarmak: @oleksii-shyman Why 'raise' is bad practice? It will just re-raise actual (the last one) exception and we can find out what the problem is. But with your variant we will make debug process much more harder I guess.

roman-verchikov: `raise` is not an issue here. 

I wonder, what if `neutron_client.create_floatingip()` raises `IpAddressGenerationFailureClient`? Do we need to handle this case as well? Is catching `NeutronClientException` too generic here maybe?

miarmak: @roman-verchikov `IpAddressGenerationFailureClient` is a child exception of `NeutronClientException`. So this case will be handled anyway. 
Another case about too generic `NeutronClientException`, that's why we check status code.

miarmak: Here are examples of this error on different client versions:
1. `IpAddressGenerationFailureClient: No more IP addresses available on network 8c6a3a43-e4a8-4e77-825b-544a080e6a7d.`
2. `NeutronClientException: 409-{u'QuantumError': u'No more IP addresses available on network d4e2bdf3-89bd-46f9-a284-66fc00a2b2f7.'}`
3. `NeutronClientException: 409-{u'NeutronError': {u'message': u"Quota exceeded for resources: ['floatingip']", u'type': u'OverQuota', u'detail': u''}}`

The one same thing they use - it is 409 status code)

roman-verchikov:
+1

toha10:
+ 1, good job, Max

PR #160, opened 2015-04-15
Fixes Bug Saic-67 : CF fails if there are no vms

oleksii-shyman:
- added if cases in 4 tasks

### DIFF
--- a/scenario/migrate.yaml
+++ b/scenario/migrate.yaml
@@ -42,6 +42,7 @@ process:
           - init_iteration_instance_copy_var: True
           - init_iteration_instance_ref: True
       - check_needed_compute_resources: True
+      - check_instances: ['rename_info_iter']
       - get_next_instance: True
       - trans_one_inst:
           - act_stop_vms: True
--- a/scenario/tasks.yaml
+++ b/scenario/tasks.yaml
@@ -36,6 +36,7 @@ tasks:
    get_next_instance: ['GetInfoIter']
    rename_info_iter: ['RenameInfo', 'info_result', 'info']
    is_instances: ['IsEndIter']
+   check_instances: ['CheckInstances']
    act_i_to_f: ['LoadComputeImageToFile', 'dst_cloud']
    act_merge: ['MergeBaseDiff', 'dst_cloud']
    act_convert_image: ['ConvertFile', 'dst_cloud']
--- a/cloud/os2os.py
+++ b/cloud/os2os.py
@@ -45,6 +45,7 @@ from cloudferrylib.os.actions import prepare_networks
 from cloudferrylib.os.actions import dissociate_floatingip_via_compute
 from cloudferrylib.os.actions import map_compute_info
 from cloudferrylib.os.actions import deploy_volumes
+from cloudferrylib.os.actions import check_instances
 from cloudferrylib.os.actions import start_vm
 from cloudferrylib.os.actions import load_compute_image_to_file
 from cloudferrylib.os.actions import merge_base_and_diff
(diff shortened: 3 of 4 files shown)


Wasapl:
Merged build finished. Test PASSed.  http://jenkins-cf.vm.mirantis.net:8080/job/cloudferry-unittests-and-style-checks/21/

roman-verchikov:
awesome, +1

miarmak:
Are you sure we need `if` statements here? Maybe it is a scenario problem and we should use here:
https://github.com/MirantisWorkloadMobility/CloudFerry/blob/devel/cloud/os2os.py#L179
the same as here:
https://github.com/MirantisWorkloadMobility/CloudFerry/blob/devel/cloud/os2os.py#L182
The same thing in the file scenario - just operate with the existing actions.
What do you think?

oleksii-shyman:
I've just checked it - it doesn't work according to wrong scenario file parser. Cloudferry actually does not start at all (because we have infinite loop). I think even if can use Your approach - we still need that ifs, because it is proper way to check if key exists before operating with it

miarmak:
Actually, it is a known issue about the scenario file parser. Known, but forgotten :) For now we should define **each** task in `scenario/tasks.yaml`, even for the same tasks. Have you tried it? Also, there shouldn't be such problem using `cloud/os2os.py` directly.

oleksii-shyman:
@miarmak  -  This approach doesn't work even in os2os.py directly

miarmak:
Yeah, that's true. Therefore it seems like problem exists in scheduler instead of file scenario parser as we thought before... 
But anyway, I think we can reuse this scheduler functionality and avoid these `if` statements with providing new task `is_not_vms`, that will be pretty similar to `is_end_iter` and make smth like this (f.e. in `os2os.py`):
`
is_not_instances = is_not_vms.IsNotVms(self.init)

        transport_instances_and_dependency_resources = \
            act_get_filter >> \
            act_get_info_inst >> \
            init_iteration_instance >> \
            act_check_needed_compute_resources >> \
            (is_not_instances | rename_info_iter) >> \
            get_next_instance >> \
            trans_one_inst >> \
            save_result >> \
            (is_instances | get_next_instance) >>\
            rename_info_iter >> \
            act_cleanup_images
        return transport_instances_and_dependency_resources
`

What do you think?

roman-verchikov:
Is this pull request still valid?

oleksii-shyman:
I think yes

oleksii-shyman:
Max, it still doesn't work.
I think We have to keep that "if" solution

Wasapl:

oleksii-shyman:
need review on that

roman-verchikov:
Looks very neat, +1

scenario/tasks.yaml
```diff
@@ -35,7 +35,9 @@ tasks:
    act_cleanup_images: ['CleanupImages']
    get_next_instance: ['GetInfoIter']
    rename_info_iter: ['RenameInfo', 'info_result', 'info']
+   rename_info_iter1: ['RenameInfo', 'info_result', 'info']
```
miarmak: Why do we need to create new instance of RenameInfo here?

scenario/migrate.yaml
```diff
@@ -42,6 +42,7 @@ process:
           - init_iteration_instance_copy_var: True
           - init_iteration_instance_ref: True
       - check_needed_compute_resources: True
+      - check_instances: ['rename_info_iter1']
```
miarmak: It looks like in this case we will not skip all transport instances tasks and will go to the next task (`get_next_instance`) if there are no instances on the source. I guess we need to do the same as in the os2os, namely use the `rename_info_iter`.

miarmak:
Overall looks good, but have a question about scenario case inline.

scenario/tasks.yaml
oleksii-shyman: because scenario parser cannot parse 2 instances in the same file

miarmak: Why do we need to make 2 instances here? We can just go forward only with 1 `rename_info_iter` here.

Wasapl:

miarmak:
Great! Thank you.
+1

roman-verchikov:
+1

PR #161, opened 2015-04-15
SAIC-131 to detect obsolete flavor ID

wzbbbb:
When a flavor is updated, the flavor id will
change. If an instance is in this flavor,
it will keep the old flavor id, can not be
matched to existing flavors
If detected, abort the migration with error
message.
	modified:   cloudferrylib/os/actions/check_needed_compute_resources.py

### DIFF
--- a/cloudferrylib/os/actions/check_needed_compute_resources.py
+++ b/cloudferrylib/os/actions/check_needed_compute_resources.py
@@ -19,6 +19,9 @@ from cloudferrylib.base import exception
 from cloudferrylib.utils import utils as utl
 
 
+LOG = utl.get_log(__name__)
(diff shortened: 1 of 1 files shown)


Wasapl:
Merged build finished. Test PASSed.  http://jenkins-cf.vm.mirantis.net:8080/job/cloudferry-unittests-and-style-checks/23/

miarmak:
Great, thanks! +1 from me

But for future, there is not necessary to have so short commit description. The recommendation is 50 characters for the first line and 72 for subsequent lines. But it is not so critical now, JFYI.

oleksii-shyman:
From my side +1

PR #166, opened 2015-04-16
Fixes SAIC-62:quantum nonetype check

oleksii-shyman:
- added simple validation

### DIFF
--- a/cloudferrylib/os/network/neutron.py
+++ b/cloudferrylib/os/network/neutron.py
@@ -818,35 +818,100 @@ class NeutronNetwork(network.Network):
     def upload_networks(self, networks):
         existing_nets_hashlist = (
             [ex_net['res_hash'] for ex_net in self.get_networks()])
+
+        # we need to handle duplicates in segmentation ids
+        # hash is used with structure {"gre": [1, 2, ...],
+        #                              "vlan": [1, 2, ...]}
+        # networks with "provider:physical_network" property added
+        # because only this networks seg_ids will be copied
+        used_seg_ids = {}
+        for net in self.get_networks():
+            if net.get("provider:physical_network"):
+                net_type = net.get("provider:network_type")
+                if net_type not in used_seg_ids:
+                    used_seg_ids[net_type] = []
+                used_seg_ids[net_type].append(
+                    net.get("provider:segmentation_id"))
+
+        networks_without_seg_ids = {}
         for net in networks:
+
             tenant_id = \
                 self.identity_client.get_tenant_id_by_name(net['tenant_name'])
+
+            # create dict, representing basic info about network
             network_info = {
-                'network':
-                {
-                    'name': net['name'],
-                    'admin_state_up': net['admin_state_up'],
+                'network': {
                     'tenant_id': tenant_id,
-                    'shared': net['shared'],
-                    'router:external': net['router:external'],
-                    'provider:physical_network': \
-                       net['provider:physical_network'],
-                    'provider:network_type':  net['provider:network_type']
+                    'admin_state_up': net["admin_state_up"],
+                    'shared': net["shared"],
+                    'name': net['name'],
+                    'router:external': net['router:external']
                 }
             }
-            if net['router:external']:
+
+            if net.get('router:external'):
                 if not self.config.migrate.migrate_extnets:
                     continue
-            if net['provider:network_type'] == 'vlan':
-                network_info['network']['provider:segmentation_id'] = (
-                    net['provider:segmentation_id'])
+            # create network on destination cloud
             if net['res_hash'] not in existing_nets_hashlist:
-                net['meta']['id'] = (self.neutron_client.
-                    create_network(network_info)['network']['id'])
+                if net.get("provider:physical_network"):
+                    # update info with additional arguments
+                    # we need to check if we have parameter
+                    # "provider:physical_network"
+                    # if we do - we need to specify 2 more
+                    # "provider:network_type" and "provider:segmentation_id"
+                    # if we don't have this parameter - creation will be
+                    # handled automatically (this automatic handling goes
+                    # after creation of networks with provider:physical_network
+                    # attribute to avoid seg_id overlap)
+                    for atr in ["provider:physical_network",
+                                "provider:network_type"]:
+                        network_info['network'].update({atr: net.get(atr)})
+
+                    # check if we have seg_ids of that type
+                    if (used_seg_ids.get(net["provider:network_type"]) and
+                        net['provider:segmentation_id'] in used_seg_ids[
+                            net["provider:network_type"]]):
+                        LOG.warning(
+                            "network {network} was dropped according"
+                            " because its segmentation id already "
+                            " exists on destination cloud".format(
+                                network=net.get("id")))
+                        continue
+                    else:
+                        if net["provider:segmentation_id"]:
+                            network_info['network'][
+                                'provider:segmentation_id'] = net[
+                                    'provider:segmentation_id']
+                        else:
+                            networks_without_seg_ids.update({
+                                net.get("id"): network_info
+                            })
+                            continue
+                    net['meta']['id'] = (
+                        self.neutron_client.create_network(
+                            network_info)['network']['id'])
+                else:
+                    # create networks later (to be sure that generated
+                    # segmentation ids don't overlap segmentation ids
+                    # created manually)
+                    networks_without_seg_ids.update({
+                        net.get("id"): network_info
+                    })
             else:
                 LOG.info("| Dst cloud already has the same network "
                          "with name %s in tenant %s" %
                          (net['name'], net['tenant_name']))
+
+        for net in networks:
+            # we need second cycle to update external object "networks"
+            # with metadata
+            if net.get("id") in networks_without_seg_ids:
+                network_info = networks_without_seg_ids[net.get("id")]
+                net['meta']['id'] = (
+                    self.neutron_client.create_network(
+                        network_info)['network']['id'])
 
     def upload_subnets(self, networks, subnets):
         existing_nets = self.get_networks()
@@ -857,10 +922,23 @@ class NeutronNetwork(network.Network):
                 continue
             tenant_id = \
                 self.identity_client.get_tenant_id_by_name(snet['tenant_name'])
+            if not tenant_id:
+                LOG.debug("Cannot get tenant_id for subnet {subnet}".format(
+                    subnet=snet.get("id")))
+                continue
             net_hash = \
                 self.get_res_hash_by_id(networks, snet['network_id'])
+            if not net_hash:
+                LOG.debug("Cannot get network info for subnet {subnet}".format(
+                    subnet=snet.get("id")))
+                continue
             network = \
                 self.get_res_by_hash(existing_nets, net_hash)
+            if not network:
+                LOG.debug("Cannot get network for subnet {subnet}".format(
+                    subnet=snet.get("id")))
+                continue
+
             subnet_info = {
                 'subnet':
                 {
@@ -897,8 +975,11 @@ class NeutronNetwork(network.Network):
             if router['external_gateway_info']:
                 ex_net_hash = \
                     self.get_res_hash_by_id(networks, router['ext_net_id'])
-                ex_net_id = \
-                    self.get_res_by_hash(existing_nets, ex_net_hash)['id']
+                ex_net = \
+                    self.get_res_by_hash(existing_nets, ex_net_hash)
+                if not ex_net:
+                    continue
+                ex_net_id = ex_net["id"]
                 r_info['router']['external_gateway_info'] = \
                     dict(network_id=ex_net_id)
             if router['res_hash'] not in existing_routers_hashlist:
--- a/tests/cloudferrylib/os/network/test_neutron.py
+++ b/tests/cloudferrylib/os/network/test_neutron.py
@@ -471,8 +471,9 @@ class NeutronTestCase(test.TestCase):
 
         self.neutron_network_client.upload_networks([self.net_1_info])
 
-        self.neutron_mock_client().create_network.\
-            assert_called_once_with(network_info)
+        if network_info['network']['provider:physical_network']:
+            self.neutron_mock_client().create_network.\
+                assert_called_once_with(network_info)
 
     def test_upload_subnets(self):
 


Wasapl:
Merged build finished. Test FAILed.   http://jenkins-cf.vm.mirantis.net:8080/job/cloudferry-unittests-and-style-checks/32/

roman-verchikov:
The number of `if`'s introduced scares the hell out of me... I wonder if there's a way we can introduce a dict-like object for client methods return values. For now it's -1 anyway, since unit tests are broken

Wasapl:
Merged build finished. Test PASSed.  http://jenkins-cf.vm.mirantis.net:8080/job/cloudferry-unittests-and-style-checks/33/

oleksii-shyman:
fixed unittests. We have a lot of different ways to fix this bug. According to our product owner - we need to find "fast" way of problem solution. I think You have to man up at not be scared of 6 ifs

roman-verchikov:
> According to our product owner - we need to find "fast" way of problem solution.
It may be fast now, but does not really fix the root cause of the problem – absence of a valid data type which tracks absence and presence of it's attributes, so hot-fixing this particular issue only (1) covers the problem; (2) requires more time to be spent in future for fixing similar issues in other modules.

> I think You have to man up at not be scared of 6 ifs
Didn't get this at all

Bottom line: +1 for now, but in future we should find a better way of transferring arguments between tasks/actions

miarmak:
I guess we started to solve this issue not from the proper side. This issue is closely related to:
1. SAIC-89
2. https://github.com/MirantisWorkloadMobility/CloudFerry/commit/33cd08183be78b055803cf7e4e5a02d7f72cf994

First of all, we need to revert changes, that were made that commit, cause it provided several new bugs.
Next, we should solve issue, mentioned in SAIC-62 about None's. And finally, solve issue with segmentation id's in SAIC-89.

And what about this specific patch (that should go after the revert changes of that commit).

This part does't need to None handling at all, cause this parameters should always be there:
https://github.com/MirantisWorkloadMobility/CloudFerry/blob/f314b9fc900a6d02a524e36722e3fafe0784a614/cloudferrylib/os/network/neutron.py#L804-L807

Another thing for these params:
https://github.com/MirantisWorkloadMobility/CloudFerry/blob/f314b9fc900a6d02a524e36722e3fafe0784a614/cloudferrylib/os/network/neutron.py#L815-L816
and these:
https://github.com/MirantisWorkloadMobility/CloudFerry/blob/f314b9fc900a6d02a524e36722e3fafe0784a614/cloudferrylib/os/network/neutron.py#L820-L821

They actually can be None. But it is not a problem for us, it can be not specified in the network at all. The problem is in providing these arguments, if they are None. If so, we just do not need to provide them to 'neutron_client.create_network'  method.

Wasapl:
Build finished. Test FAILed.   http://jenkins-cf.vm.mirantis.net:8080/job/cloudferry-unittests-and-style-checks/36/

cfglib.py
```diff
@@ -107,6 +107,8 @@
                help='Path to file with the groups of VMs'),
     cfg.BoolOpt('all_networks', default=False,
                 help="Migrate all network resources from all tenants"),
+    cfg.BoolOpt('drop_duplicated_seg_ids', default=True,
+                help="Migrate all network resources from all tenants"),
```
roman-verchikov: Update help message with description of `drop_duplicated_seg_ids` option.

Wasapl:
Merged build finished. Test FAILed.   http://jenkins-cf.vm.mirantis.net:8080/job/cloudferry-unittests-and-style-checks/37/

Wasapl:
Merged build finished. Test PASSed.  http://jenkins-cf.vm.mirantis.net:8080/job/cloudferry-unittests-and-style-checks/38/

cloudferrylib/os/network/neutron.py
```diff
@@ -818,31 +818,74 @@ def upload_sec_group_rules(self, sec_groups):
...
             [ex_net['res_hash'] for ex_net in self.get_networks()])
+
+        # we need to handle duplicates in segmentation ids
+        used_seg_ids = [i.get('provider:segmentation_id') for i in networks]
+
+        def get_free_segmentation_id():
+            # this function finds first free seg_id
+            # adds it to array of used ids
+            # and returns seg_id
+            for i in xrange(1, 4096):
```
roman-verchikov: This will only work for VLANs and only if `network_vlan_ranges` neutron config option is set to 1:4096. Ranges for other segmentation types are different:

 * VXLANs
 ** (24-bit segmentation ID)[http://tools.ietf.org/html/rfc7348#section-5], meaning ranges from 0 to (2^24)-1;
 ** `vni_ranges` config option in neutron specifies the actual range used in cloud
 * GREs do not require segmentation IDs to be identical on source and destination since it's a tunneling protocol, it only requires the numbers to be set for a particular tenant.

With that in mind, I guess the correct way would be:
```
if segmentation_type in [ "VLAN", "VXLAN" ]:
    segmentation_id = <SRC_SEGMENTATION_ID>
    for _ in number_of_retries:
        try:
            neutron net-create --segmentation-id= <netname>
            break
        except SegmentationIdConflict:
            segmentation_id += 1
            continue
elif segmentation_type == "GRE":
    neutron net-create <netname>
```

oleksii-shyman: according to https://github.com/openstack/neutron/blob/stable/icehouse/neutron/plugins/ml2/drivers/type_tunnel.py#L77
GRE requires segmentation id (I get message NeutronClientException: 400-{u'NeutronError': {u'message': u'Invalid input for operation: segmentation_id required for gre provider network.', u'type': u'InvalidInput', u'detail': u''}})

Also - I don't like try except approach, because it will send more then one request per network, Also I don't think that we need that "number of retries", because we can lose data if we exceed number of retries. But I will find a way to handle different number of possible segmentation ids.

roman-verchikov: > according to https://github.com/openstack/neutron/blob/stable/icehouse/neutron/plugins/ml2/drivers/type_tunnel.py#L77
GRE requires segmentation id (I get message NeutronClientException: 400-{u'NeutronError': {u'message': u'Invalid input for operation: segmentation_id required for gre provider network.', u'type': u'InvalidInput', u'detail': u''}})

This is only true for provider networks (when user wants to handle real physical network through Openstack network)

> Also - I don't like try except approach, because it will send more then one request per network, Also I don't think that we need that "number of retries", because we can lose data if we exceed number of retries. But I will find a way to handle different number of possible segmentation ids.

Only a suggestion on how to track segmentation IDs

Wasapl:
Merged build finished. Test PASSed.  http://jenkins-cf.vm.mirantis.net:8080/job/cloudferry-unittests-and-style-checks/40/

Wasapl:
Merged build finished. Test PASSed.  http://jenkins-cf.vm.mirantis.net:8080/job/cloudferry-unittests-and-style-checks/41/

oleksii-shyman:
need review on this pull req

roman-verchikov:
Looks ok, but we should seriously consider having different more generalized approach on these kinds of issues. Otherwise we'll end up verifying every single dict key prior to using it.

cloudferrylib/os/network/neutron.py
```diff
@@ -818,36 +818,95 @@ def upload_sec_group_rules(self, sec_groups):
...
+                            LOG.warning(
+                                "network {network} was dropped according"
+                                " because its segmentation id already "
+                                " exists on destination cloud".format(
+                                    network=net.get("id")))
+                            continue
+                    else:
+                        seg_id = net['provider:segmentation_id']
+                    network_info['network']['provider:segmentation_id'] = (
+                        seg_id)
```
miarmak: It seems like we can have a situation, when `seg_id ` variable is not defined.

cloudferrylib/os/network/neutron.py
```diff
@@ -857,10 +916,23 @@ def upload_subnets(self, networks, subnets):
...
+                continue
             net_hash = \
                 self.get_res_hash_by_id(networks, snet['network_id'])
+            if not net_hash:
+                LOG.debug("Cannot get network info for subnet {subnet}".format(
+                    subnet=snet.get("id")))
+                continue
             network = \
                 self.get_res_by_hash(existing_nets, net_hash)
+            if not network:
```
miarmak: I guess we do not need this check, because we can not create a subnet without a network. It is mandatory requirement.

cloudferrylib/os/network/neutron.py
```diff
@@ -857,10 +916,23 @@ def upload_subnets(self, networks, subnets):
                 continue
             tenant_id = \
                 self.identity_client.get_tenant_id_by_name(snet['tenant_name'])
+            if not tenant_id:
```
miarmak: The same as below. We can not create a subnet outside of any tenant.

cloudferrylib/os/network/neutron.py
```diff
@@ -818,36 +818,95 @@ def upload_sec_group_rules(self, sec_groups):
     def upload_networks(self, networks):
         existing_nets_hashlist = (
             [ex_net['res_hash'] for ex_net in self.get_networks()])
+
+        # we need to handle duplicates in segmentation ids
+        # hash is used with structure {"gre": [1, 2, ...],
+        #                              "vlan": [1, 2, ...]}
+        # networks with "provider:physical_network" property added
+        # because only this networks seg_ids will be coppied
```
miarmak: Typo in word 'copied'

cloudferrylib/os/network/neutron.py
```diff
@@ -818,36 +818,95 @@ def upload_sec_group_rules(self, sec_groups):
...
-                net['meta']['id'] = (self.neutron_client.
-                    create_network(network_info)['network']['id'])
+                if net.get("provider:physical_network"):
+                    # update info with additional arguments
+                    # we need to check if we have parameter
+                    # "provider:physical_network"
+                    # if we do - we need to specify 2 more
+                    # "provider:network_type" and "provider:segmentation_id"
+                    # if we don't have this parameter - creation will be
+                    # handled automaticaly (this automatic handling goes
```
miarmak: Typo in word 'automatically'

cloudferrylib/os/network/neutron.py
```diff
@@ -818,36 +818,95 @@ def upload_sec_group_rules(self, sec_groups):
...
+                    # after creation of networks with provider:physica_network
```
miarmak: typo in 'provider:physical_network' attribute

cloudferrylib/os/network/neutron.py
```diff
@@ -818,36 +818,95 @@ def upload_sec_group_rules(self, sec_groups):
...
+                        seg_id = net['provider:segmentation_id']
+                    network_info['network']['provider:segmentation_id'] = (
+                        seg_id)
+                    net['meta']['id'] = (
+                        self.neutron_client.create_network(
+                            network_info)['network']['id'])
+                else:
+                    # create networks later (to be sure that generated
+                    # segmentation ids don't overlap segmentation ids
+                    # created manualy)
```
miarmak: Typo in word 'manually'

miarmak:
And also, unfortunately it doesn't work for me:

`INFO     [/Users/maksym/GitHub/CloudFerry/cloudferrylib/scheduler/scheduler.py : lineno 47]  04-29 15:55:18  -------- Start task: BaseTask|NetworkTransporter
ERROR    [/Users/maksym/GitHub/CloudFerry/cloudferrylib/scheduler/scheduler.py : lineno 58]  04-29 15:56:02  BaseTask|NetworkTransporter TASK FAILED
Traceback (most recent call last):
  File "/Users/maksym/GitHub/CloudFerry/cloudferrylib/scheduler/scheduler.py", line 71, in process_chain
    self.run_task(task)
  File "/Users/maksym/GitHub/CloudFerry/cloudferrylib/scheduler/scheduler.py", line 63, in run_task
    self.map_func_task[task](task)
  File "/Users/maksym/GitHub/CloudFerry/cloudferrylib/scheduler/scheduler.py", line 92, in task_run
    task(namespace=self.namespace)
  File "/Users/maksym/GitHub/CloudFerry/cloudferrylib/scheduler/task.py", line 96, in __call__
    result = self.run(**namespace.vars)
  File "/Users/maksym/GitHub/CloudFerry/cloudferrylib/os/actions/networks_transporter.py", line 26, in run
    dst_resource.deploy(info)
  File "/Users/maksym/GitHub/CloudFerry/cloudferrylib/os/network/neutron.py", line 75, in deploy
    self.upload_networks(deploy_info['networks'])
  File "/Users/maksym/GitHub/CloudFerry/cloudferrylib/os/network/neutron.py", line 889, in upload_networks
    network_info)['network']['id'])
  File "/Users/maksym/GitHub/CloudFerry/cloudferrylib/utils/proxy_client.py", line 78, in __call__
    raise e
NeutronClientException: Request Failed: internal server error while processing your request.
INFO     [/Users/maksym/GitHub/CloudFerry/cloudferrylib/scheduler/scheduler.py : lineno 76]  04-29 15:56:02  Failed processing CHAIN MIGRATION

Done.`

cloudferrylib/os/network/neutron.py
miarmak: Ok, here it can be. Sorry

miarmak: But anyway, I don't think that we need such checks here.

cloudferrylib/os/network/neutron.py
oleksii-shyman: that's why we have this check. if method    self.identity_client.get_tenant_id_by_name(snet['tenant_name'])    doesn't return valid data - we continue

cloudferrylib/os/network/neutron.py
oleksii-shyman: we do need that check, because self.get_res_by_hash(existing_nets, net_hash) can return None

Wasapl:

cloudferrylib/os/network/neutron.py
miarmak: ok

cloudferrylib/os/network/neutron.py
miarmak: That's true. But from another side we can not check all such cases, cause our code will be grown up with huge amount of `if's`. Even in this module we have quite a few such situations, where we do not do checks...

miarmak:
And unfortunately I still have the same error as above.

oleksii-shyman:
I cannot reproduce this error

Wasapl:

miarmak:
Looks good and works for me. +1
But I would suggest @toha10 to review this patch.

Wasapl:

toha10:
Guys, do you really need to keep network special settings during the migration?

Usually we have clouds with different ranges of segmentation ids, f.e. SRC cloud can have 1000:1100 ids range and destination 2000:3000 ids range. Also they may have different network types (vlan, gre, flat). These settings strongly depend from network backend which is used by neutron service and underlying network environment.
If we not keep this special settings during the migration, new networks will get next available id from the pool.
Keeping this settings needed if we migrate external networks, but in this case sometimes we do not need to keep these settings, because src and dst clouds also may have different network structure, f.e. external network on SRC cloud was flat and connected to physnet1, but on DST cloud we will need to create ext net as vlan network and we need to connect it to physnet 3. For this kind of differences  i think we need to add an additional file with special parameters for network.
F.e.:
external network id on SRC cloud = dict of new parameters

roman-verchikov:
> Guys, do you really need to keep network special settings during the migration?

@toha10, yes, in our case we must transfer all the data to DST exactly as it was on SRC, including low-level network segmentation IDs.

oleksii-shyman:
@toha10  -  I think that for now we have to keep current implementation, in case we need to extend functionality - we have to provide that mapping solution

miarmak:
I'm okay with it, if it does not break anything. For now as I see on our environments (dev and lab), it seems to work fine.

PR #173, opened 2015-04-23
Changes in configuration template and tests import

plomakin:
Added lost parameters for networks, vms, images.
Fixed imports for tests from load_generator.

### DIFF
--- a/devlab/config.template
+++ b/devlab/config.template
@@ -20,6 +20,9 @@ ssh_chunk_size = 100
 retry = 5
 migrate_extnets = True
 time_wait = 5
+all_vms = False
+all_images = True
+all_networks = True
 
 [src]
 type = os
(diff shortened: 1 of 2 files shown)


Wasapl:
Merged build finished. Test PASSed.  http://jenkins-cf.vm.mirantis.net:8080/job/cloudferry-unittests-and-style-checks/56/

miarmak:
+1

devlab/config.template
```diff
@@ -20,6 +20,9 @@ ssh_chunk_size = 100
 retry = 5
 migrate_extnets = True
 time_wait = 5
+all_vms = False
```
roman-verchikov: Don't we want this to be set to `True` by default?

plomakin: I believe it is False in CF direcory: configs/config.ini
So, i've just copied it to template as is.

roman-verchikov:
+1

PR #179, opened 2015-04-30
SAIC-170 Start ssh-agent once for each private key

roman-verchikov:
ssh-agent now starts for the first ssh command issued and never gets
killed. This is done in order to improve end-user (deployment engineer)
experience in case he/she has password-protected private key. With this
patch enabled password is only required to be entered once for the whole
migration process.

### DIFF
--- a/cloudferrylib/utils/utils.py
+++ b/cloudferrylib/utils/utils.py
@@ -13,7 +13,6 @@
 # limitations under the License.
 
 import logging
-import sys
 import time
 import timeit
 import random
@@ -28,6 +27,7 @@ import os
 import inspect
 from multiprocessing import Lock
 from fabric.api import run, settings, local, env
+from fabric.context_managers import hide
 import ipaddr
 import yaml
 from logging import config
@@ -296,8 +296,7 @@ def log_step(log):
     return decorator
 
 
-class forward_agent:
-
+class forward_agent(object):
     """
         Forwarding ssh-key for access on to source and destination clouds via ssh
     """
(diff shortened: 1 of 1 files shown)


Wasapl:

cloudferrylib/utils/utils.py
```diff
@@ -296,27 +296,45 @@ def inner(*args, **kwargs):
...
         self.ssh_auth_sock = info_agent[1]
         os.environ["SSH_AGENT_PID"] = self.pid
         os.environ["SSH_AUTH_SOCK"] = self.ssh_auth_sock
 
     def __exit__(self, type, value, traceback):
-        local("kill -9 %s"%(self.pid))
-        del os.environ["SSH_AGENT_PID"]
-        del os.environ["SSH_AUTH_SOCK"]
+        # never kill previously started ssh-agent, so that user only has to
+        # enter private key password once
```
miarmak: Does it mean, that in the end of migration we will have ssh-agent launched?

miarmak:
Looks good and works for me. Have little question inline.

cloudferrylib/utils/utils.py
roman-verchikov: Yes, exactly that. And I don't see any problem with that. Actually I think ssh-agent should be started prior to running cloudferry.

Wasapl:

cloudferrylib/utils/utils.py
miarmak: Ok

miarmak:
+1

PR #181, opened 2015-05-04
SAIC-167: Fix bug with nova security groups

miarmak:
Sometimes in get_instance response key "security_groups" is absent.
This patch makes safe receiving of security_groups with nova's response.


roman-verchikov:
Kind of OK, but this is one more workaround for the items not present in the object. There will be a zillion more of these unless we fix the root of the issue. +1 for now

Wasapl:

miarmak:
@roman-verchikov It is not our fault, it is Nova's issue. Here is for example the same thing in Manila:
https://bugs.launchpad.net/manila/+bug/1320938

PR #182, opened 2015-05-05
Fix ssh keys insertion in box-repackage script

roman-verchikov:
Public keys are inserted correctly, but since VMs are shutdown
immediately after the filesystem modification, the data is not stored.
Call `sync` to fix that.


roman-verchikov:
Since this patch is completely unrelated to core functionality, it can be merged without review.

PR #185, opened 2015-05-06
SAIC-156: Fix DHCP issue for all tenants

miarmak:
Since the creation order of DHCP port and instance port can be random,
sometimes, the DHCP port can take the IP already used by instance on the
source could. To avoid it, this fix will disable DHCP for all networks
in all tenants, participated im migration process, and free it's IP
before instance port creation, then re-enable the DHCP.

### DIFF
--- a/cloudferrylib/os/actions/prepare_networks.py
+++ b/cloudferrylib/os/actions/prepare_networks.py
@@ -31,12 +31,17 @@ class PrepareNetworks(action.Action):
         keep_ip = self.cfg.migrate.keep_ip
 
         instances = info_compute[utl.INSTANCES_TYPE]
+
+        # Get all tenants, participated in migration process
+        tenants = set()
+        for instance in instances.values():
+            tenants.add(instance[utl.INSTANCE_BODY]['tenant_name'])
+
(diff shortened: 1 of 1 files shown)


Wasapl:

cloudferrylib/os/actions/prepare_networks.py
```diff
@@ -31,12 +31,17 @@ def run(self, info=None, **kwargs):
         keep_ip = self.cfg.migrate.keep_ip
 
         instances = info_compute[utl.INSTANCES_TYPE]
+
+        # Get all tenants, participated in migration process
+        tenants = set([])
```
oleksii-shyman: why do we need that brackets?
set() works

oleksii-shyman:
looks ok + 1

cloudferrylib/os/actions/prepare_networks.py
miarmak: Just a habit of declaring the empty set) Will fix it.

Wasapl:

roman-verchikov:
+1

PR #186, opened 2015-05-06
Minor: Get rid of 'get_security_groups_list' method

miarmak:
Since this method is unused, obsolete and duplicate - it has been simply
deleted from cloudferrylib/os/network/neutron.py.


roman-verchikov:
Less code is better code, +1

Wasapl:

PR #187, opened 2015-05-06
SAIC-190 Fix VM migration for VMs with ERROR state

roman-verchikov:
In case SRC has VMs with ERROR state, migration procedure fails accessing
the compute host where the VM is expected to be run. This patch fixes
the issue by verifying the VM is not in error state.


Wasapl:

oleksii-shyman:
Looks ok + 1

miarmak:
For me too. +1

PR #188, opened 2015-05-08
SAIC-169 Key pair migration

roman-verchikov:
Key pairs are only accessible by user created them, admin user has no
access to user key pairs. Because of that migration fails for non-admin
VMs which have keypair associated.

### DIFF
--- a/cloudferrylib/os/actions/transport_compute_resources.py
+++ b/cloudferrylib/os/actions/transport_compute_resources.py
@@ -11,16 +11,18 @@
 # implied.
 # See the License for the specific language governing permissions and#
 # limitations under the License.
-
-
 import copy
 
 from cloudferrylib.base.action import action
+from cloudferrylib.os.compute import keypairs
+from cloudferrylib.os.identity import keystone
 from cloudferrylib.utils import utils as utl
 
 
-class TransportComputeResources(action.Action):
+LOG = utl.get_log(__name__)
 
+
+class TransportComputeResources(action.Action):
     def run(self, info=None, identity_info=None, **kwargs):
         info = copy.deepcopy(info)
         target = 'resources'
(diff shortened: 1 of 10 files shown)


Wasapl:

cloudferrylib/os/actions/transport_compute_resources.py
```diff
@@ -38,3 +40,144 @@ def run(self, info=None, identity_info=None, **kwargs):
...
+            dst_nova.mysql_connector.execute(sql)
+
+    def run(self, **kwargs):
+        src_keystone = self.src_cloud.resources[utl.IDENTITY_RESOURCE]
+        dst_keystone = self.dst_cloud.resources[utl.IDENTITY_RESOURCE]
+
+        LOG.info('Fetching users present on SRC and DST, this may take awhile')
+
+        src_users = src_keystone.get_users_list()
+        dst_users = dst_keystone.get_users_list()
```
roman-verchikov: It's definitely a very bad idea to fetch all users, there can be hundreds of thousands of them in corporate LDAP in general case. I will rewrite this to only get user ID which has key pair on SRC

Wasapl:

Wasapl:

roman-verchikov:
Ok, so the problem is actually more complicated than I thought. Simply put, it's not enough to transfer key pairs to destination on the low level, it is also required to boot VM on destination with original user credentials, in our case VM is booted with admin credentials at all times, which makes it impossible to associate VM with user's key pair (key pair is only visible to user, not to admin). I will continue investigating the possibility to hack cloud-init on the lower level so that it provides correct keys at boot time without `--key-pair` argument provided to `nova boot`.

Wasapl:
Merged build finished. Test FAILed.

Wasapl:

Wasapl:

Wasapl:

plomakin:
+1 Works good for simple/devlab scenarios

Wasapl:

PR #189, opened 2015-05-12
SAIC-207: Change Glance client authentication process

miarmak:
In some configurations there are no availability to retrieve OpenStack
services list via Keystone. But Glance authentication was completely
dependent on the services manipulations via Keystone API. That's why we
need to change the way of the Glance authentication. From now it gets
public endpoint URL directly using service type (image).

Also, 'get_endpoint_by_service_type' method was added to Keystone
resource.

### DIFF
--- a/cloudferrylib/os/image/glance_image.py
+++ b/cloudferrylib/os/image/glance_image.py
@@ -68,8 +68,10 @@ class GlanceImage(image.Image):
     def get_client(self):
         """ Getting glance client """
 
-        endpoint_glance = self.identity_client.get_endpoint_by_service_name(
-            'glance')
+        endpoint_glance = self.identity_client.get_endpoint_by_service_type(
+            service_type='image',
+            endpoint_type='publicURL')
+
         # we can figure out what version of client to use from url
         # check if we have "v1" or "v2" in the end of url
         m = re.search("(.*)/v(\d)", endpoint_glance)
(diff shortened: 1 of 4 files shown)


Wasapl:

wzbbbb:
This fixed the issue.

cloudferrylib/os/image/glance_image.py
```diff
@@ -68,8 +68,10 @@ def get_db_connection(self):
     def get_client(self):
         """ Getting glance client """
 
-        endpoint_glance = self.identity_client.get_endpoint_by_service_name(
```
roman-verchikov: Looks like `get_endpoint_by_service_name` method is not used anywhere else, lets remove it

roman-verchikov:
I wonder if we should consider leaving both methods of finding keystone endpoints for compatibility reasons. It might happen that some clouds only support searching endpoints by name, not type.

miarmak:
@roman-verchikov I am not sure we need both cases. Searching via service type is more preferable way. And also I've tested it on 3 different clouds - everything was fine.

roman-verchikov:
> @roman-verchikov I am not sure we need both cases. Searching via service type is more preferable way. And also I've tested it on 3 different clouds - everything was fine.

Ok, let's then simply remove `get_endpoint_by_service_name`, it's not used.

cloudferrylib/os/image/glance_image.py
miarmak: Done

miarmak:
DO NOT MERGE

Wasapl:

Wasapl:

roman-verchikov:
+1, seem to be working, according to @wzbbbb

PR #195, opened 2015-05-15
External networks mapping option

toha10:
Added external networks mapping option (ext_net_map in config file).
This option allow to map external networks on source cloud to
existing external networks on destination cloud with appropriate parameters by
dictionary with old and new ids.

### DIFF
--- a/cloudferrylib/os/network/neutron.py
+++ b/cloudferrylib/os/network/neutron.py
@@ -37,6 +37,8 @@ class NeutronNetwork(network.Network):
         self.cloud = cloud
         self.identity_client = cloud.resources['identity']
         self.neutron_client = self.proxy(self.get_client(), config)
+        self.ext_net_map = \
+            utl.read_yaml_file(self.config.migrate.ext_net_map) or {}
 
     def get_client(self):
         return neutron_client.Client(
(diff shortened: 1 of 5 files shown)


Wasapl:
Merged build finished. Test FAILed.

Wasapl:

cloudferrylib/os/network/neutron.py
```diff
@@ -37,6 +37,8 @@ def __init__(self, config, cloud):
         self.cloud = cloud
         self.identity_client = cloud.resources['identity']
         self.neutron_client = self.proxy(self.get_client(), config)
+        self.ext_net_map = utl.read_yaml_file(self.config.migrate.ext_net_map) \
+            if utl.read_yaml_file(self.config.migrate.ext_net_map) else {}
```
roman-verchikov: `self.ext_net_map = utl.read_yaml_file(self.config.migrate.ext_net_map) or {}` would be much cleaner.

roman-verchikov:
Would be great if you could isolate external networks logic into separate class and cover it with unit tests. At the very least let's have the introduced logic being tested at least.

cloudferrylib/os/network/neutron.py
toha10: Thank you for the notice

Wasapl:

toha10:
Speaking about external networks logic - it's just mapping that is used in case if we have already created external networks on dest cloud and we want to use these networks instead of migration the old ones.
Could you please clarify: are you proposing to isolate the whole external networks getting logic ?

roman-verchikov:
> are you proposing to isolate the whole external networks getting logic ?

Yeah, but that's probably too much of an effort, so let's just cover introduced logic with unit tests for now.

plomakin:
+1 doesn't affect common functionality

roman-verchikov:
+1

roman-verchikov:
Two +1s, can be merged

PR #205, opened 2015-05-27
SAIC-215: Move SSH credentials to config

miarmak:
- Add availability to specify source and destination users and `sudo`
passwords for them in configs/config.ini;
- Change 'direct_compute_transfer' logic. From now this option suppose,
that CloudFerry node has direct connection to both source and destination
compute nodes and also they have connectivity with each other;
- Fix and extend `SSHChunksTransfer` driver;
- Set `SSHChunksTransfer` driver as a default ssh file transfer driver
for `transport_ephemeral` task if `direct_compute_transfer` is enabled;
- Add new task `act_trans_diff_file_chunks` to scenario/tasks.yaml and
use it as default in scenario/migrate.yaml for instance diff file transfer;
- Add `get_remote_file_size` method to utils;
- Fix some PEP8 issues in these files.


Wasapl:
Merged build finished. Test PASSED.

cfglib.py
```diff
@@ -23,7 +23,11 @@
     cfg.StrOpt('temp', default='-',
                help='temporary directory on controller'),
     cfg.StrOpt('service_tenant', default='services',
-               help='Tenant name for services')
+               help='Tenant name for services'),
+    cfg.StrOpt('ssh_user', default='root',
+               help='user to connect via ssh'),
+    cfg.StrOpt('ssh_sudo_password', default='',
```
roman-verchikov: I think default password should be `None`, I would assume "no password" is different from "empty password".

cfglib.py
```diff
@@ -50,7 +54,11 @@
...
```
roman-verchikov: Same as above

cloudferrylib/utils/utils.py
```diff
@@ -507,3 +516,16 @@ def import_class_by_string(name):
...
+def get_remote_file_size(host, file_path, ssh_user, ssh_sudo_password):
+    """ Return file size in bytes on the remote host.
+
+    :param host: Remote host,
+    :param file_path: Full file path,
+    :return: File size in bytes.
+    """
+
+    with settings(host_string=host, user=ssh_user, password=ssh_sudo_password):
+        size = run('wc -c < %s' % file_path)
```
roman-verchikov: Not sure if it's safe to retrieve file size with `wc`:
 - it might not work correctly for binary files
 - in this particular case it will count the number of *bytes* in file, which is not always equal to the file size.

I would recommend using `stat` instead.

scenario/migrate.yaml
```diff
@@ -79,7 +79,8 @@ process:
               - act_deploy_instances: True
               - add_key_pairs_to_instances: True
               - act_is_not_copy_diff_file: ['act_transport_ephemeral']
-              - act_trans_diff_file: True
+              - act_trans_diff_file: False
+              - act_trans_diff_file_chunks : True
```
roman-verchikov: Let's also update migration scenario in `devlab/tests/migrate.yaml`

cloudferrylib/os/actions/transport_ephemeral.py
```diff
@@ -20,17 +20,9 @@
...
-from cloudferrylib.os.actions import copy_g2g
 from cloudferrylib.os.actions import task_transfer
-from cloudferrylib.utils import utils as utl, forward_agent
-
-from cloudferrylib.utils.drivers import ssh_ceph_to_ceph
-from cloudferrylib.utils.drivers import ssh_ceph_to_file
-from cloudferrylib.utils.drivers import ssh_file_to_file
-from cloudferrylib.utils.drivers import ssh_file_to_ceph
+from cloudferrylib.utils import forward_agent
+from cloudferrylib.utils import utils as utl
```
roman-verchikov: do we really need to save 2 bytes here and sacrifice readability?

cfglib.py
miarmak: Actually, as we already have discussed, there is no difference for fabric.

cloudferrylib/os/actions/transport_ephemeral.py
miarmak: No, we don't need it actually) It is just 'legacy' from line 28 :)

scenario/migrate.yaml
miarmak: Ok

cloudferrylib/utils/utils.py
miarmak: Ok, will change.

Wasapl:
Merged build finished. Test PASSED.

scenario/migrate.yaml
```diff
@@ -18,7 +18,7 @@ process:
           - act_get_info_images_src: True
           - act_get_info_inst_src: True
           - act_get_info_volumes_src: True
-          - act_get_info_objects_src: True
+          - act_get_info_objects_src: False
```
roman-verchikov: `act_get_info_objects_src` and `act_get_info_objects_dst` seem to be unrelated to the patch

miarmak: Oh yeah, it is left after testing. Will remove it.

Wasapl:
Merged build finished. Test PASSED.

plomakin:
+1, Nice job, dude!

roman-verchikov:
+1, looks good. Created a ticket to refactor ssh_chunks.py logic separately.

PR #206, opened 2015-05-27
SAIC-235 Fix condensation for flavors with non-int IDs

roman-verchikov:
Condensation used to fail in case flavor was non-int value. This
patch fixes it.

### DIFF
--- a/condensation/flavor.py
+++ b/condensation/flavor.py
@@ -18,8 +18,7 @@ class Flavor(object):
     """This class is representation of OpenStack flavor"""
 
     def __init__(self, fl_id, name, ram, core, ephemeral=None, swap=None):
-        # by some reason - id of flavor is string
-        self.fl_id = int(fl_id)
+        self.fl_id = fl_id
         self.name = name
         self.ram = ram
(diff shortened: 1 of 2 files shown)


Wasapl:
Merged build finished. Test PASSED.

condensation/flavor.py
```diff
@@ -18,8 +18,7 @@ class Flavor(object):
     """This class is representation of OpenStack flavor"""
 
     def __init__(self, fl_id, name, ram, core, ephemeral=None, swap=None):
-        # by some reason - id of flavor is string
-        self.fl_id = int(fl_id)
```
miarmak: Don't you know any reason to have it in `int` type before?

roman-verchikov: There's no reason, all openstack objects define IDs as string.

miarmak: Okay. Just trying to find out the reason of this type conversion)

miarmak:
Looks good! +1

Wasapl:
Merged build finished. Test PASSED.

PR #215, opened 2015-06-05
Small fixes in config template for devlab.

plomakin:
Fixed path to public key.
Fixed temp paths on SRC and DST clouds.

PR #217, opened 2015-06-06
Fix bug with temporary directories path

miarmak:
Fix binding to slash ('/') in the path of temporary directories, defined
by user in the config. So, from now there is a possibility to specify
there path either with the slash in the end of temp dir path
(f.e. '/tmp/') and without it (f.e. '/tmp').

### DIFF
--- a/cloudferrylib/os/actions/check_bandwidth.py
+++ b/cloudferrylib/os/actions/check_bandwidth.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 
+import os
 import subprocess
 import uuid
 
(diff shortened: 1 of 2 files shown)


miarmak:
Also fixes #215 issue.

Wasapl:
Build finished.
Unittest success

Wasapl:
Unittest success

Wasapl:
Build finished.
Functional test success

Wasapl:
Functional test success

roman-verchikov:
Looks ok, would be better to use `mktemp` probably, but this should work too. +1

PR #235, opened 2015-06-23
SAIC-270: Implement custom DB settings for compute

miarmak:
Not always CloudFerry node has remote root access to the source and/or
destination clouds databases. And not always all OpenStack services
databases are located on the same host. This patch provides possibility
to specify custom settings (driver, host, user, password, database name)
to the Nova compute databases for both src and dst clouds.

### DIFF
--- a/cfglib.py
+++ b/cfglib.py
@@ -211,7 +211,17 @@ src_compute_opts = [
     cfg.BoolOpt('block_migration', default=False,
                 help='live-migration without shared_storage'),
     cfg.StrOpt('host_eph_drv', default='-',
-               help='host ephemeral drive')
+               help='host ephemeral drive'),
+    cfg.StrOpt('connection', default='mysql+mysqlconnector',
+               help='driver for db connection'),
+    cfg.StrOpt('host', default='',
+               help='compute mysql node ip address'),
+    cfg.StrOpt('database_name', default='',
+               help='compute database name'),
+    cfg.StrOpt('user', default='',
+               help='user for db access'),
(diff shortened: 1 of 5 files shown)


Wasapl:
Build finished.
Unittest success

Wasapl:
Unittest success

Wasapl:
Build finished.
Functional test success

Wasapl:
Functional test success

cfglib.py
```diff
@@ -211,7 +211,17 @@
     cfg.BoolOpt('block_migration', default=False,
                 help='live-migration without shared_storage'),
     cfg.StrOpt('host_eph_drv', default='-',
-               help='host ephemeral drive')
+               help='host ephemeral drive'),
+    cfg.StrOpt('connection', default='mysql+mysqlconnector',
+               help='driver for db connection'),
+    cfg.StrOpt('host', default='',
```
roman-verchikov: I would recommend renaming option names to have `db_` prefix, for example, `db_host` instead of `host`, `db_user` instead of `user` and so on. I would also recommend renaming the rest of DB options in other config groups according to that pattern, it is much more readable, IMHO.

miarmak: It is definitely good idea, but there are a lot of name changes we should do in the whole project (OpenStack custom db settings, general db settings) and do not break everything. In this case, I would suggest to leave everything as it is, create a 'nice to have' issue with this renaming and do it as a separate refactoring task someday. What do you think?

roman-verchikov:
Looks good, works on dev and prod, can be merged.

PR #246, opened 2015-06-30
Box repackage script creates public key if it doesn't exist

roman-verchikov:
The code in box repackage script relied on having public key stored in
vagrant directory. This patch removes this dependency and generates
public key from private key in case public key does not exist.


roman-verchikov:
Very minor patch not touching core functionality, was verified and tested with icehouse v0.0.14 box. Can be merged.

Wasapl:
Build finished.
Functional test failure

Wasapl:
Functional test failure

PR #268, opened 2015-07-10
Minor fix: Add SSH options to devlab config

miarmak:
According to SAIC-215, new SSH settings (ssh user and ssh sudo password)
have been provided to CloudFerry, but they are missing in the devlab
config. This patch fixes the issue.


vnaboychenko:
"if" statement looks good for me

realsystem:
+1, looks good

devlab/provision/generate_config.sh
```diff
@@ -34,7 +34,7 @@ while read key value
 do
     value=($value)
     value=${value[1]}
-    if [[ -n ${value} ]]; then
+    if [[ -n ${value} ]] || [[ "${value}" == "" ]]; then
```
roman-verchikov: This would work for all values – both empty and non-empty. Simply remove if

roman-verchikov:
+1

devlab/provision/generate_config.sh
```diff
@@ -34,9 +34,7 @@ while read key value
 do
     value=($value)
     value=${value[1]}
-    if [[ -n ${value} ]]; then
-      sed -i "s|<${key}>|${value}|g" ${result_config}
-    fi
+    sed -i "s|<${key}>|${value}|g" ${result_config}
```
realsystem: Looks like after this operation we will have key without value in the configurationfile. I think it is possible to remove key line without value. Like this:
    if [[ -n ${value} ]]; then
        sed -i "s|<${key}>|${value}|g" ${result_config}
    elif [[ -z ${value} && -n ${key} ]]; then
        sed -i "/$key/d" ${result_config}
    fi

miarmak: But what is the reason to remove it? I thought the purpose of this patch was to add it instead...

realsystem: remove if empty, add if full

Settis: No! We must save it empty.
For example we have `prop` with default value `defVal`

Config file line | result
------------------|---------
prop=someVal | someVal
prop= | 
 | defVal

And if you delete this line from config file you will have default value.

realsystem: As I understand default value should be set when user's value is not set. So, when user keep a blank key - it means that there is no value and need to use default value to prevent any mistakes. Let's see comments from other reviewers.

roman-verchikov: +1 to @miarmak and @Settis. I think this can be merged.

miarmak: @realsystem, When user keep a blank key - it means that there will be an empty value (`''`). When user does not specify key at all - it means that there will be default value.

roman-verchikov:
Still need one more +1 to merge

realsystem:
+1

roman-verchikov:
can be merged

PR #269, opened 2015-07-10
[SAIC-266] CF fails when try to filter by non-existing VM id

Settis:
When you try to run CF with non-existing VM id in filter config it will fail with:
> NotFound: Instance could not be found (HTTP 404) (Request-ID: req-5cb28e1c-8bab-40ac-ba57-29a96e475791)

This pull makes the error more useful:
>  NotFound: No server with ID of '<some id>' exists.

### DIFF
--- a/cloudferrylib/os/compute/nova_compute.py
+++ b/cloudferrylib/os/compute/nova_compute.py
@@ -504,10 +504,13 @@ class NovaCompute(compute.Compute):
                 detailed=detailed, search_opts=search_opts, marker=marker,
                 limit=limit)
         else:
-            if type(ids) is list:
-                instances = [self.nova_client.servers.get(i) for i in ids]
-            else:
-                instances = [self.nova_client.servers.get(ids)]
+            ids = ids if type(ids) is list else [ids]
+            instances = []
+            for i in ids:
+                try:
+                    instances.append(self.nova_client.servers.get(i))
+                except nova_exc.NotFound:
+                    LOG.warning("No server with ID of '%s' exists." % i)
 
         instances = filter_down_hosts(
             down_hosts(self.get_client()), instances,


cloudferrylib/os/compute/nova_compute.py
```diff
@@ -504,10 +505,14 @@ def get_instances_list(self, detailed=True, search_opts=None, marker=None,
...
-                instances = [self.nova_client.servers.get(i) for i in ids]
-            else:
-                instances = [self.nova_client.servers.get(ids)]
+            ids = ids if type(ids) is list else [ids]
+            instances = []
+            for i in ids:
+                try:
+                    instances.append(self.nova_client.servers.get(i))
+                except nova_exc.NotFound:
+                    raise cf_exc.NotFound("No server with ID of '%s' exists."
```
roman-verchikov: Would it be better to simply ignore these cases instead of raising an error?

miarmak: I guess we need to skip this error with printing in the Warning log. At least, if we want to have it here. Otherwise, we need to have it on the filter level. Or we can implement universal checking action for such purposes, cause currently we also have images filtering too, and in future there will be more different objects (f.e. we plan add volumes filtering).

cloudferrylib/base/exception.py
```diff
@@ -19,3 +19,6 @@ class OutOfResources(Exception):
 
 class ImageDownloadError(Exception):
     pass
+
```
miarmak: PEP8 violation: Add please 1 more blank line.

Settis:
Only warning log was implemented for instances.
If we start filtering images by non existing id we don't catch NotFound exception. Take a look at code
``` python
    def get_image_by_id(self, image_id):
        for glance_image in self.get_image_list():
            if glance_image.id == image_id:
                return glance_image
```

cloudferrylib/os/compute/nova_compute.py
```diff
@@ -22,6 +22,7 @@
 
 import cfglib
 from cloudferrylib.base import compute
+from cloudferrylib.base import exception as cf_exc
```
roman-verchikov: This is not used, please remove

roman-verchikov:
Looks good other than 1 inline.

Settis:
Done.

roman-verchikov:
+1

realsystem:
Need to check it by functional tests

miarmak:
Seems to be working:
http://jenkins-cf.vm.mirantis.net:8080/job/cloudferry-functional-tests/555/

+1.

PR #272, opened 2015-07-14
Use exception handler in remote runner to handle fabric ssh errors

realsystem:


cloudferrylib/utils/remote_runner.py
```diff
@@ -52,10 +53,13 @@ def run(self, cmd):
...
-                    return sudo(cmd)
-                else:
-                    return run(cmd)
+                try:
+                    if self.sudo:
+                        return sudo(cmd)
+                    else:
+                        return run(cmd)
+                except RemoteExecutionError:
+                    raise RuntimeError('Wrong credentials for %s' % self.host)
```
vnaboychenko: what if remote command return non-zero status?

PR #280, opened 2015-07-17
Updated scripts for cleaning lab, deploy lab and run migration

realsystem:
Added all-tenants support for os_cli image cleaning.
Added usage description for each script.

### DIFF
--- a/devlab/utils/clean_run.sh
+++ b/devlab/utils/clean_run.sh
@@ -8,7 +8,25 @@
 set -e
 
 cur_pwd=`pwd`
-cfg=devlab/config.ini
+cfg="devlab/config.ini"
+dftl_ip="192.168.1.2"
+
+usage() {
+    echo
+    echo "Script to run clean migration"
+    echo "It uses $cfg for getting access to environment"
+    echo "If you want to do it for default options, please, use -f"
+    echo "  bash $0 -f"
+    echo "In other case the script uses $cfg"
+    echo "Make sure you have update it properly for all IPs"
+    echo
+}
+
+if [[ "$1" != "-f" ]] && [[ `grep -q $dftl_ip $cfg && echo $?` == "0" ]]; then
+    echo "Warning! Default config detected"
+    usage
+    exit 1
+fi
 
 SRC=`grep grizzly_ip $cfg | awk '{print $3}'`
 DST=`grep icehouse_ip $cfg | awk '{print $3}'`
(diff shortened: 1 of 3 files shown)


devlab/utils/clean_run.sh
```diff
@@ -34,7 +34,8 @@ python generate_load.py --clean --env DST
 python generate_load.py --clean --env DST
 ssh root@$DST \
 "mysql -u $dst_mysql_user -p$dst_mysql_password cinder -e \"delete from cinder.volume_admin_metadata;
-delete from cinder.volumes;\""
+delete from cinder.volumes;delete from cinder.reservations;delete from cinder.quota_usages;\""
```
miarmak: You can use all tables from here: https://github.com/MirantisWorkloadMobility/CloudFerry/blob/devel/cloudferrylib/os/storage/cinder_database.py#L53-L61

devlab/utils/clean_run.sh
```diff
@@ -34,7 +34,8 @@ python generate_load.py --clean --env DST
...
 
 cd $cur_pwd
+bash devlab/utils/os_cli.sh clean dst || echo "Can't clean dst"
```
miarmak: Are we gonna start migration even if 2nd condition is met?

miarmak:
Also, commit message is too long) Reduce it please to 50 symbols. 
And also, there are a lot of different unrelated changes in commit, so you could list it below in the commit description (<72 symbols). :wink:

devlab/utils/README.md
```diff
@@ -0,0 +1,16 @@
...
+    cd CloudFerry
+    change IPs in devlab/config.ini
+    bash devlab/utils/os_cli.sh clean src
+    and/or
+    bash devlab/utils/os_cli.sh clean dst
+
+3. Clean environment and run migration
+    cd CloudFerry
+    change IPs in devlab/config.ini
+    bash devlab/utils/clean_run.sh
```
roman-verchikov: Readme does not seem very useful, because each script already has documentation in the header. Looks like unnecessary double work in the future to support both.

If you believe we need it:
1. Please use markdown syntax for code (```)
2. Let's limit line length to 120 symbols
3. There's also `repackage_box.sh` script in this folder let's have description for it in README too.

miarmak: I would suggest to have documentation in one place.

devlab/utils/clean_run.sh
Settis: If you want to delete all data from table `truncate` is better.
Also you can select all table names from mysql:
`"mysql -u $dst_mysql_user -p$dst_mysql_password -Nse 'show tables' cinder | while read table; do mysql -u $dst_mysql_user -p$dst_mysql_password -e \"set foreign_key_checks=0; truncate table $table\" cinder; done"`

roman-verchikov: > Also you can select all table names from mysql
I guess the order is important ;)

roman-verchikov:
+1

miarmak:
Looks good. +1

devlab/utils/clean_run.sh
Settis: I agree with you. Especially for that I add `set foreign_key_checks=0`.

PR #286, opened 2015-07-21
[SAIC-300] Add Nova default quotas migration

miarmak:
Implement availability to migrate Nova Compute default quotas. It is
possible if `migrate_quotas` option of config's `migrate` section is
`True`.


realsystem:
+1

roman-verchikov:
+1, can  be merged

PR #289, opened 2015-07-23
[SAIC-349] Fix `skip_down_hosts` for `check_ssh`

miarmak:
Add new `get_compute_hosts` method to the Nova Compute resource and use
it in the `check_ssh` action. This was done because we actually need to
operate computes in this check, but not hypervisors.
Also, `get_hypervisors` method has been deleted from `nova_compute`
resource, because it is not used anywhere else in the project.


roman-verchikov:
Looks good, +1

realsystem:
+1

PR #290, opened 2015-07-24
SAIC-356 Add admin user to non-admin tenant

roman-verchikov:
Implements class which allows safe operations on objects which do not
belong to admin user. Currently required for VM to be booted and
floating IP to be created in non-admin tenant.


realsystem:
+1

miarmak:
Looks good. +1

roman-verchikov:
Got two +1s, can be merged

PR #301, opened 2015-07-29
SAIC-337 Same networks have different hashes

roman-verchikov:
In case network on source has incorrectly defined CIDR, migrated network
will have CIDR changed to correct, which will break the network
comparison.

Example:
 - source net CIDR: 64.143.228.11/22
 - migrated net CIDR: 64.143.228.0/22

In the given example networks are the same, yet CF incorrectly compares
equality of these networks by their string representation.

This patch fixes this behavior.

### DIFF
--- a/cloudferrylib/os/network/neutron.py
+++ b/cloudferrylib/os/network/neutron.py
@@ -14,6 +14,7 @@
 import pprint
 
 import ipaddr
+import netaddr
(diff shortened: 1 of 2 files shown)


miarmak:
Looks good. +1

realsystem:
+1

PR #302, opened 2015-07-29
SAIC-125: added script for filter preparation

realsystem:


devlab/jenkins/prep_filter.sh
```diff
@@ -0,0 +1,23 @@
...
+
+filter="configs/filter.yaml"
+cfg="devlab/config.ini"
+
+src=$(grep grizzly_ip $cfg | awk '{print $3}')
+ssh_user=$(grep src_ssh_user $cfg | awk '{print $3}')
+ssh_cmd="ssh $ssh_user@${src}"
+
+if [[ $1 == "--tenant" ]]; then
+a="awk '{print \$2}'"
```
roman-verchikov: Is it here just to follow the 79 symbols per line rule?.. It's quite unreadable, also shell scripts don't have to follow pep8, I think we can have 120 symbols for non-python code.

devlab/jenkins/prep_filter.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
+
+set -e
+
+tenant_name="tenant1"
```
roman-verchikov: What if `tenant1` is not available? Let's pass tenant name as argument to the script

roman-verchikov:
+1

PR #307, opened 2015-07-30
SAIC-370 Split floating IP associate/disassociate into different tasks

roman-verchikov:
Separates floating IP associate/disassociate from port creation. This is required for the live migration to have minimal downtime during floating IP association for migrated instance and disassociation of the same floating IP on source.


scenario/tasks.yaml
```diff
@@ -10,7 +10,9 @@ tasks:
    act_stop_vms: ['StopVms', 'src_cloud']
    act_start_vms: ['StartVms', 'dst_cloud']
    act_start_vms_if_needed: ['StartVmsIfNeeded', 'dst_cloud']
-   act_dissociate_floatingip: ['DissociateFloatingip', 'src_cloud']
+   act_dissociate_floatingip: ['DisassociateAllFloatingips', 'src_cloud']
+   disassociate_floatingip_on_source: ['DisassociateFloatingip', 'src_cloud']
```
realsystem: Can you explain naming of tasks here? One is named act_* but another named without act_

roman-verchikov: I decided not to change the naming of the original task, while for the new one it seems more readable to have it named `disassociate_floatingip_on_source`

realsystem:
+1 and re-run func tests, it should be ok after 2-3 tries :)

PR #310, opened 2015-07-31
[SAIC-371] Fix issue with obsolete images

miarmak:
Remove information about obsolete/broken images (and snapshots) from
`info` variable, because we do not migrate them. We need to delete it,
cause this information is used during the migration process (for example
for images mapping) and can break the migration.

Also, logging has been removed from the `get_image_ref` method of
`glance_image` resource due to it's uselessness.

### DIFF
--- a/cloudferrylib/os/image/glance_image.py
+++ b/cloudferrylib/os/image/glance_image.py
@@ -134,9 +134,7 @@ class GlanceImage(image.Image):
     def get_ref_image(self, image_id):
         try:
             return self.glance_client.images.data(image_id)._resp
-        except exc.HTTPInternalServerError as e:
-            LOG.error("There is an error occurred while trying to reach "
-                      "image's data. (id = %s)\n %s", image_id, e)
+        except exc.HTTPInternalServerError:
             raise exception.ImageDownloadError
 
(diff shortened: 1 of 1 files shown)


cloudferrylib/os/image/glance_image.py
```diff
@@ -135,7 +135,7 @@ def get_ref_image(self, image_id):
         try:
             return self.glance_client.images.data(image_id)._resp
         except exc.HTTPInternalServerError as e:
-            LOG.error("There is an error occurred while trying to reach "
+            LOG.debug("There is an error occurred while trying to reach "
```
roman-verchikov: Hmmm... this looks like error, let's simply remove this log from here, and log actual message there.

miarmak: Yeah, it is. But the main point here was to not see this error message in the output and logs (cause it has big scary red traceback, that can confuse someone) during the migration process. But instead of it, we will be able to see this information if we enable Debug mode.

miarmak:
Fixed, tested and ready for review.

realsystem:
+1

roman-verchikov:
+1

PR #328, opened 2015-08-11
Decreased RAM for Vagrant VMs

mgrygoriev:
It was found by testing that there is enough 2,5GB of RAM for Source and Destination VMs and 1,5Gb for CloudFerry VM.


roman-verchikov:
looks good

PR #329, opened 2015-08-11
SAIC-402 Support for passwords with special characters

roman-verchikov:
Refactors direct file copy logic, and adds support for passwords with special characters.

### DIFF
--- a/cloudferrylib/utils/drivers/ssh_chunks.py
+++ b/cloudferrylib/utils/drivers/ssh_chunks.py
@@ -14,11 +14,7 @@
 
 
 import os
-
-from fabric.api import env
-from fabric.api import hide
-from fabric.api import run
-from fabric.api import settings
+import math
 
 from cloudferrylib.utils import driver_transporter
 from cloudferrylib.utils import files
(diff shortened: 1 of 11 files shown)


roman-verchikov:
Seem to be failing with unrelated issue: ```Missing resources on dest: set([u'm1.medium', u'm1.small', u'm1.xlarge', u'm1.large'])")```

cloudferrylib/utils/files.py
```diff
@@ -89,4 +90,19 @@ def _unlink_dir(dirname):
...
+
+
+class RemoteTempDir(object):
+    """Creates remote temp dir using `mktemp` and removes it on scope exit"""
+
+    def __init__(self, runner):
+        self.runner = runner
+
+    def __enter__(self):
+        create_temp_dir = 'mktemp -d'
```
realsystem: We already have GetTempDir to get directory name, it can be used here. Or use the same format of directory "mktemp -dt %s_XXXX" % self.prefix

tests/cloudferrylib/utils/test_file_utils.py
```diff
@@ -57,3 +59,81 @@ def test_temp_dir_is_deleted_on_scope_exit(self):
...
+
+        self.assertEqual(actual_iterations, expected_iterations)
+
+
+class VerifiedFileCopyTestCase(test.TestCase):
+    @mock.patch("tests.cloudferrylib.utils.test_file_utils.ssh_chunks."
+                "remote_scp")
+    @mock.patch("tests.cloudferrylib.utils.test_file_utils.ssh_chunks."
+                "remote_md5_sum")
+    def test_raises_error_on_copy_failure(self, scp, _):
```
realsystem: Can you, please, explain who send "scp" to this test?

cloudferrylib/utils/files.py
roman-verchikov: This one is different, because it creates dir, not only returns possible remote path. This is safer (no race conditions) and faster (single call needed for creation)

tests/cloudferrylib/utils/test_file_utils.py
roman-verchikov: It's mock of `ssh_chunks.remote_scp`. See this to understand why I patched this instead of `cloudferrylib.utils.drivers.ssh_chunks`.: http://www.voidspace.org.uk/python/mock/patch.html#where-to-patch

cloudferrylib/utils/files.py
realsystem: Ok, what about the format of directory name? After this patch we will have different format. I think it's better to use one format like in GetTempDir. If we don't need prefix - make it the same in all places in the code.

cloudferrylib/utils/drivers/ssh_chunks.py
```diff
@@ -29,177 +25,178 @@
...
-                                    (ssh_user_dst, host_dst,
-                                     'echo \'%s\' | sudo -S %s' % ('<password>',
-                                                                   command)))
-
-    def _calculate_parts_count(self, data):
-        part_size = self.cfg.migrate.ssh_chunk_size
-
-        byte_size = data.get(
-            'byte_size',
-            utils.get_remote_file_size(data['host_src'],
```
realsystem: Looks like get_remote_file_size is not used anymore, can be deleted. It uses run directly and this is terrible. Many issues can be fixed by using runner.run.

PR #330, opened 2015-08-11
SAIC-301 test for volume migration with non-existing user

ntymtsiv:


devlab/tests/generate_load.py
```diff
@@ -426,6 +433,12 @@ def modify_admin_tenant_quotas(self):
                     'admin'), **tenant['quota'])
                 break
 
+    def delete_users(self):
+        for user in config.users:
+            if 'deleted' in user and user['deleted']:
```
roman-verchikov: `user.get('deleted')` is 2 times faster ;)

PR #333, opened 2015-08-11
SAIC-405 Fix nova flavor migration

roman-verchikov:
The issue was caused by invalid novaclient.flavors.get() behavior, which
returns deleted flavors if flavor ID matches. Replaced with
novaclient.flavors.find(), which raises 404 `NotFound` if flavor is
deleted.


roman-verchikov:
Can be merged.

PR #334, opened 2015-08-12
SAIC-400 Rollback for VM in 'verify resize' state

roman-verchikov:
Fixes issue when rolling back failed migration fails itself with `KeyError`.
The issue appeared because there was no transition from 'shutdown' to
'verify_resize' state defined in code.


roman-verchikov:
Tested locally, works fine

PR #335, opened 2015-08-13
Tweak vagrant settings for VMs

mgrygoriev:
Decreased CPU usage limit from 100% to 80% and increased RAM for icehouse VM to 3G


roman-verchikov:
+1

PR #336, opened 2015-08-13
SAIC-390 Schedule VM boot host manually on error

roman-verchikov:
Picks random host if nova-scheduled compute node fails to boot VM. This is
useful in case one of the nodes in cluster is displayed to nova as operational,
but is faulty in reality.

### DIFF
--- a/cloudferrylib/os/compute/nova_compute.py
+++ b/cloudferrylib/os/compute/nova_compute.py
@@ -14,6 +14,7 @@
 
 
 import copy
+import random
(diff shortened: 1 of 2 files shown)


roman-verchikov:
```Bandwidth is not OK. Claimed bandwidth: 100 Mb/s. Required speed: 50.0 Mb/s. Actual upload speed: 56.97 Mb/s. Actual download speed: 43.48 Mb/s. Aborting migration..```

I think this check must be removed

PR #337, opened 2015-08-13
Non-public flavor migration unit test fix

roman-verchikov:
Fixed incorrect unit test for non-public nova flavor migration.


roman-verchikov:
Has no impact on functionality, can be merged.

PR #338, opened 2015-08-13
Decrease speed_limit in config.template

mgrygoriev:
This change need to be done to prevent CI system to fail builds because
of slow network connectivity between VMs


roman-verchikov:
+1

roman-verchikov:
Looks like the option modified is not used in `CheckBandwidth`. You should have changed `[initial_check] factor` option instead.

mgrygoriev:
Fixed

roman-verchikov:
+1

PR #340, opened 2015-08-14
SAIC-406: Fixed IOError in generate_load

ntymtsiv:


roman-verchikov:
-1.
Solution only works if `generate_load.py` is started from `CloudFerry` root dir, and fails for any other folder. The solution must be generic, so that user can run `generate_load.py` from anywhere.

ntymtsiv:
I have re-tested and it works from anywhere. We get CloudFerry root dir respectively to file devlab/tests/filtering_utils.py, not from dir where generate_load was started.

roman-verchikov:
yep, seem to be working, can be merged.

PR #345, opened 2015-08-17
Quick start guide

roman-verchikov:
Adds quick start guide document for new engineers willing to start
playing with CloudFerry.

### DIFF
--- a/QUICKSTART.md
+++ b/QUICKSTART.md
@@ -1,0 +1,59 @@
+# Quick Start Guide
+
+1. Get CloudFerry sources
+    ```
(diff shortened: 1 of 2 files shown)


realsystem:
Looks like a duplicate of README. I think we need one place for this info.

roman-verchikov:
> realsystem commented a day ago
> Looks like a duplicate of README. I think we need one place for this info.

These files really have some intersections (like dependency installation), yet the purpose is different: README provides general information about the tool, while QUICKSTART gives direct instructions on how to run the simplest migration scenario which works 100%. This seems to be quite useful

PR #346, opened 2015-08-18
Fixed vms filtering in generate load

ntymtsiv:


devlab/tests/generate_load.py
```diff
@@ -437,13 +437,15 @@ def emulate_vm_states(self):
...
         data = {}
-        for vm in self.novaclient.servers.list():
-            vm_name = vm.name
-            index = self.novaclient.servers.list().index(vm)
-            while self.novaclient.servers.list()[index].status == u'RESIZE':
-                time.sleep(2)
-            vm_state = self.novaclient.servers.list()[index].status
-            data[vm_name] = vm_state
+        for vm in self.novaclient.servers.list(search_opts={'all_tenants': 1}):
+            for i in range(600):
```
realsystem: It's better to make definition at the top of file with this value(600).

realsystem:
+1, except one comment above

devlab/tests/generate_load.py
roman-verchikov: Magic numbers are considered bad practice in general. Please refactor to something readable.

devlab/tests/functional_test.py
```diff
@@ -31,3 +41,52 @@ class FunctionalTest(unittest.TestCase):
     def __init__(self, *args, **kwargs):
         super(FunctionalTest, self).__init__(*args, **kwargs)
         suppress_dependency_logging()
+        self.src_cloud = Prerequisites(cloud_prefix='SRC')
+        self.dst_cloud = Prerequisites(cloud_prefix='DST')
+        self.filtering_utils = FilteringUtils()
+
+    def filter_resources(self, obj):
```
roman-verchikov: **Much** better would be to have a separate `filter` method for each type of object, any method **must not** deal with objects of different type. This solution not only complicates implementation, it's also completely unreadable, hardly maintainable and difficult to read.

roman-verchikov:
please rebase

devlab/tests/functional_test.py
```diff
@@ -31,3 +34,97 @@ class FunctionalTest(unittest.TestCase):
...
+        return [i for i in self.src_cloud.glanceclient.images.list()
+                if i.name in images]
+
+    def filter_volumes(self):
+        volumes = [i['name'] for i in config.cinder_volumes]
+        opts = {'search_opts': {'all_tenants': 1}}
+        return [i for i in self.src_cloud.cinderclient.volumes.list(**opts)
+                if i.display_name in volumes]
+
+    def __get_neutron_resources(self, res, names):
```
roman-verchikov: You know of the fact double underscores in the beginning of attribute name have [special meaning](https://docs.python.org/2/tutorial/classes.html#private-variables-and-class-local-references) in python, don't you? Make sure this is something you really need.

roman-verchikov:
please rebase too

devlab/tests/functional_test.py
ntymtsiv: I used double underscores to mark methods as private

roman-verchikov: Private method must be marked with single underscore. Methods started with two underscores are treated in a special way: https://docs.python.org/2/tutorial/classes.html#private-variables-and-class-local-references

roman-verchikov:
Please check why functional tests are failing and fix comment inline.

roman-verchikov:
+1, can be merged.

PR #350, opened 2015-08-19
SAIC-419 Remove unneeded endpoint definition

roman-verchikov:
Removes glance and neutron endpoints definition from generate_load.py


realsystem:
+1

roman-verchikov:
Looks like this patch doesn't help in production (endpoints are not listable there) and in devlab (migration stops). Closing for now.

PR #351, opened 2015-08-19
Modify vagrant VMs' deleting procedure

mgrygoriev:
We delete only VMs created by current Jenkins build


roman-verchikov:
looks good, but still it killed the other job running in parallel :(

roman-verchikov:
looks good, let's see how parallel builds go with it

PR #359, opened 2015-08-27
Fixed data_collector script, to collecect only required data.

VitaliiYerys:
Data collector will not collect info about when VM was updated,
because it will be updated anyway during rollback, thats why this
data was excluded from collected_data dict in order to avoid errors.


roman-verchikov:
+1, can be merged

PR #360, opened 2015-08-27
Modified generate_load to use config dynamically

ntymtsiv:


roman-verchikov:
please rebase

roman-verchikov:
Please add description to pull request and to commit message, it's not clear what this patch is about from the title.

devlab/tests/generate_load.py
```diff
@@ -270,14 +270,14 @@ def get_vm_nics(_vm):
...
+        for vm in self.config.vms:
             vm['image'] = self.get_image_id(vm['image'])
             vm['flavor'] = self.get_flavor_id(vm['flavor'])
             vm['nics'] = get_vm_nics(vm)
             self.check_vm_state(self.novaclient.servers.create(**vm))
-        for tenant in config.tenants:
-            if 'vms' in tenant:
-                for user in config.users:
+        for tenant in self.config.tenants:
+            if tenant.get('vms'):
```
roman-verchikov: Minor: this modification is not really needed, you could leave the old code here as it's actually more readable that this.

devlab/tests/generate_load.py
```diff
@@ -563,8 +563,12 @@ def clean_router_ports(router_id):
         def delete_user_keypairs(_user):
             if not _user.get('enabled'):
                 return
-            self.switch_user(user=_user['name'], password=_user['password'],
-                             tenant=_user['tenant'])
+            try:
+                self.switch_user(user=_user['name'],
+                                 password=_user['password'],
+                                 tenant=_user['tenant'])
+            except Exception:
```
roman-verchikov: Catching `Exception` is considered bad practice, can you make exception handling here more narrow? Also, please add `log` or `print`, so that at least the exception is analyzable. 

Simple example of why it's bad idea to catch `Exception`: in case `_user` does not have `'tenant'` key for whatever reason (someone modified `config.py` and forgot about it), in current code you'll simply ignore that error, and will get further through the code and finally get even more cryptic error message, or (which is even worse) invalid results.

devlab/tests/generate_load.py
```diff
@@ -642,29 +650,39 @@ def wait_until_vms_all_deleted():
...
                     snapshot['image_name'], repr(e))
-        for user in config.users:
+
+        sgs = self.neutronclient.list_security_groups()['security_groups']
+        for sg in sgs:
+            try:
+                print "delete sg {}".format(sg['name'])
+                self.neutronclient.delete_security_group(self.get_sg_id(
+                                                         sg['name']))
+            except Exception as e:
```
roman-verchikov: Same as above. Please consider narrowing down the list of exceptions you want to handle here (maybe `neutronclient.exceptions.BaseException`?..)

roman-verchikov:
+1

PR #367, opened 2015-09-01
Restart glance during icehouse VM provisioning

mgrygoriev:


devlab/jenkins/functional_tests.sh
```diff
@@ -17,7 +17,11 @@ clean_exit()
     pushd ${CF_DIR}/devlab
     vboxmanage list vms
     vagrant status
-    vboxmanage list vms | grep "${BUILD_NAME:1:${#BUILD_NAME}}" | awk -F'_' '{print $2}' | xargs -I {} vagrant destroy {} --force
+    vboxmanage list vms | grep "${BUILD_NAME:1:${#BUILD_NAME}}"
+    if [ $? -ne 0 ]; then
```
roman-verchikov: Bash will never get here if `$?` is non-zero, `set -e` will make it exit before

devlab/Vagrantfile
```diff
@@ -83,6 +86,7 @@ Vagrant.configure(2) do |config|
 
   nodes.each do |nodename, nodedata|
     config.vm.define nodename do |thisnode|
+      thisnode.vm.network "forwarded_port", guest: 22, host: nodedata['sshport'], id: 'ssh'
```
roman-verchikov: `sshport` may be empty for user

roman-verchikov:
Looks ok, but we'd have to split devlab from jenkins. This task is beyond this patch though

PR #369, opened 2015-09-02
[SAIC-423] ClouldFerry fails migration if target subnet on dst exists

Settis:
Network overlapping check was added into `preparation` section in the end of `pre_migration_test`.

### DIFF
--- a/cloudferrylib/os/actions/check_networks.py
+++ b/cloudferrylib/os/actions/check_networks.py
@@ -1,0 +1,98 @@
+# Copyright (c) 2015 Mirantis Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the License);
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an AS IS BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or
+# implied.
+# See the License for the specific language governing permissions and#
+# limitations under the License.
+
+import ipaddr
+import collections
+import exceptions
+from cloudferrylib.base.action import action
+from cloudferrylib.utils import utils
+
+
+LOG = utils.get_log(__name__)
+
+
+class CheckNetworks(action.Action):
+    """
+    Check subnets overlapping and raise EnvironmentError if needed.
+    It must be done before actual migration in check section.
+    The action use filter search opts and must run after act_get_filter action
+    """
+
+    def run(self, **kwargs):
+        LOG.debug("Start 'check_networks' action")
+        src_net = self.src_cloud.resources[utils.NETWORK_RESOURCE]
+        dst_net = self.dst_cloud.resources[utils.NETWORK_RESOURCE]
+        search_opts = kwargs.get('search_opts_tenant', {})
+        src_info = NetworkInfo(src_net.read_info(**search_opts))
+        dst_info = NetworkInfo(dst_net.read_info(**search_opts))
+        for network in src_info.get_networks():
+            dst_net = dst_info.by_hash.get(network.hash)
+            if dst_net:
+                LOG.debug("src network %s, dst network %s" %
+                          (network.id, dst_net.id))
+                network.check_network_overlapping(dst_net)
+
+
+class NetworkInfo(object):
+    def __init__(self, info):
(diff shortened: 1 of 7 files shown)


cloudferrylib/os/actions/check_networks.py
```diff
@@ -0,0 +1,85 @@
...
+import collections
+import exceptions
+from cloudferrylib.base.action import action
+from cloudferrylib.utils import utils
+
+
+LOG = utils.get_log(__name__)
+
+
+class CheckNetworks(action.Action):
```
roman-verchikov: Please add docstring describing
 - What it does
 - How and when to use
 - Dependencies, dependents and configuration options (if needed)

cloudferrylib/os/actions/check_networks.py
```diff
@@ -0,0 +1,85 @@
...
+    def run(self, **kwargs):
+        LOG.debug("Start 'check_netwroks' action")
```
roman-verchikov: 'networks' is misspelt in check_netwroks

cloudferrylib/os/actions/check_networks.py
```diff
@@ -0,0 +1,85 @@
...
+        src_net = self.src_cloud.resources[utils.NETWORK_RESOURCE]
+        dst_net = self.dst_cloud.resources[utils.NETWORK_RESOURCE]
+        search_opts = kwargs.get('search_opts_tenant', {})
+        src_info = NetworkInfo(src_net.read_info(**search_opts))
+        dst_info = NetworkInfo(dst_net.read_info(**search_opts))
+        for network in src_info.get_networks():
```
roman-verchikov: Please consider refactoring to make the process more linear, it has too many levels of indentation right now (I counted 7).

cloudferrylib/os/network/neutron.py
```diff
@@ -279,7 +279,6 @@ def convert_subnets(snet, cloud):
         }
 
         res_hash = network_res.get_resource_hash(result,
-                                                 'name',
```
roman-verchikov: This doesn't seem right to me right now. Why are we not comparing name of a network?..

devlab/tests/scenarios/cold_migrate.yaml
```diff
@@ -26,6 +26,7 @@ preparation:
           - check_dst_sql: True
           - check_dst_rabbit: True
           - act_get_info_objects_dst: False
+      - check_networks: True
```
roman-verchikov: Please add to live_migrate.yaml too

cloudferrylib/os/network/neutron.py
Settis: I think subnet name meaningless when we already have the same subnet on dst except name.
In this way if someone manually creates subnet on dst with needed `cidr`, gateway, ect, we will use it.

roman-verchikov: Yes, but having "name" here gives more chances our subnet comparison is correct, because tenant can have more than one subnet with the same CIDR. For example having name will handle following case correctly, while without name it will be invalid:
```
# as tenant t1
neutron net-create net1
neutron subnet-create --name subnet1 net1 10.0.0.0/24
neutron net-create net2
neutron subnet-create --name subnet1 net2 10.0.0.0/24
```

Settis: Yes. You are right.
What about adding `network_name` instead of subnet `name`?

tests/cloudferrylib/os/actions/test_check_networks.py
```diff
@@ -0,0 +1,129 @@
...
+
+from tests import test
+from cloudferrylib.os.actions import check_networks
+
+
+class CheckNetworksTestCase(test.TestCase):
+
+    def setUp(self):
+        super(CheckNetworksTestCase, self).setUp()
+        check_networks.LOG.setLevel("DEBUG")
```
roman-verchikov: Please remove this, logs are not expected to be present in unit tests.

cloudferrylib/os/network/neutron.py
roman-verchikov: Having `network_name` is good, but I still think we should also have subnet `name` too, it will narrow the number of possible false alarms.

tests/cloudferrylib/os/actions/test_check_networks.py
Settis: Done

roman-verchikov:
+1

PR #371, opened 2015-09-03
Fixed SAIC-387: refactoring for wait_for_status

realsystem:
wait_for_status is used from base class for images, volumes and instances.

### DIFF
--- a/cloudferrylib/base/resource.py
+++ b/cloudferrylib/base/resource.py
@@ -12,7 +12,12 @@
 # See the License for the specific language governing permissions and#
 # limitations under the License.
 
+import time
 from cloudferrylib.utils import proxy_client
+from cloudferrylib.utils import timeout_exception
+from cloudferrylib.utils import utils
+
+LOG = utils.get_log(__name__)
 
 
 class Resource(object):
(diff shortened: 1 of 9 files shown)


cloudferrylib/base/resource.py
```diff
@@ -36,8 +41,18 @@ def save(self):
     def restore(self):
         pass
 
-    def wait_for_status(self, id_obj, status, limit_retry=60):
-        pass
+    def wait_for_status(self, res_id, get_status, wait_status, limit_retry=60, warn_only=False):
```
roman-verchikov: pep8 violation - line too long

roman-verchikov: Even though logic follows earlier versions of the code, I think it is not fully correct. Wait for status should wait no longer than configured value (say, 5 minutes), and raise error or print log, it should not rely on number of retries. Something like
```
def wait_for_status(openstack_object, expected_status)
    delay = 1
    while delay < config.timeout:
        if openstack_object_status != expected_status:
            break
        sleep(delay)
        delay *= 2
    else:
        raise Timeout()
```

Also, instead of handling raise or not raise an error as part of a single method, let's better have two different methods which work differently:
```
def wait_for_status()
"""Raises error"""
    <implementation>

def try_wait_for_status()
"""Not raises error"""
    try:
        wait_for_status()
    except Timeout as e:
        LOG.warning(e)
```

roman-verchikov: Don't forget to cover the functionality with unit tests

realsystem: #run pep checker
find . -name '*.py' -not -wholename './.venv/*' | xargs .venv/bin/flake8 --max-line-length 160 || true

roman-verchikov: > find . -name '.py' -not -wholename './.venv/' | xargs .venv/bin/flake8 --max-line-length 160 || true

Not funny. Max line length is [79 according to pep8](https://www.python.org/dev/peps/pep-0008/#maximum-line-length)

roman-verchikov:
+1, nice!

PR #374, opened 2015-09-03
[SAIC-438] Templating of creating a database connection resource

mirrorcoder:


cloudferrylib/os/network/neutron.py
```diff
@@ -41,6 +41,7 @@ def __init__(self, config, cloud):
         self.filter_tenant_id = None
         self.ext_net_map = \
             utl.read_yaml_file(self.config.migrate.ext_net_map) or {}
+        self.mysql_connector = cloud.mysql_connector('neutron')
```
roman-verchikov: This may fail for pre-havana releases, where neutron was called "quantum".

roman-verchikov:
looks good other than one inline, please fix

mirrorcoder:
I changed logic for mysql config in configuration.ini, base config is section src_mysql/dst_mysql another sections will override mysql credentional

cloudferrylib/os/network/neutron.py
mirrorcoder: if we need another db_name for network then we can change in configuration.ini

roman-verchikov:
please rebase

roman-verchikov:
+1

PR #383, opened 2015-09-07
Fixed SAIC-437: added new pre-migration checks

realsystem:


roman-verchikov:
I think we need to extend these checks to verify we can create and delete an openstack object on destination cloud. The primary reason for this is to make sure destination cloud is functional, this is something we often have problems with.

realsystem:
If we add such check we will "change" DST before creating snapshot/dump. So, after that if something goes wrong in this new check - we will make "bad" DST.
I think new check should be placed in "process" block in scenario.

cloudferrylib/os/actions/check_openstack_apis.py
```diff
@@ -19,11 +19,14 @@
...
 
 
-def check(os_api_call, os_api_type, position):
+def check(os_api_call, os_api_type, position, arg=None):
     try:
         LOG.info("Checking %s APIs availability on %s.",
                  os_api_type, position.upper())
-        os_api_call()
+        if arg is not None:
+            os_api_call(arg)
```
roman-verchikov: let's simply pass `args` and `kwargs` without any `if`s, e.g.:
```
def check(os_api_call, os_api_type, position, *os_api_call_args, **os_api_call_kwargs):
    ...
    # no ifs here
    os_api_call(*os_api_call_args, **os_api_call_kwargs)
    ...
```

roman-verchikov:
> If we add such check we will "change" DST before creating snapshot/dump. So, after that if something goes wrong in this new check - we will make "bad" DST.
> I think new check should be placed in "process" block in scenario.
I agree, but let's have these changes in this pull request too.

cloudferrylib/os/actions/check_openstack_apis.py
```diff
@@ -19,11 +19,14 @@
...
+        else:
+            os_api_call()
     except Exception as e:
```
roman-verchikov: Let's also limit the exceptions caught here to all base exceptions of openstack clients, e.g.:
```
from neutronclient import exceptions as neutron_excs
from glanceclient import exceptions as glance_excs
...
try:
    ...
except (neutron_excs.BaseException, glance_excs.BaseException, ...) as e:
    ...
```

roman-verchikov:
+1

roman-verchikov:
can be merged

PR #393, opened 2015-09-11
SAIC-448 Added test 'image which belongs to deleted tenant migration'

ntymtsiv:


devlab/tests/config.py
```diff
@@ -1,3 +1,5 @@
+img_url = 'http://download.cirros-cloud.net/0.3.3/cirros-0.3.3-x86_64-disk.img'
```
realsystem: Do you need a real image for testing migration? As I know you can touch an empty file and use it for glance. Or just use raw data in API. Please, see https://github.com/MirantisWorkloadMobility/CloudFerry/pull/383/files for reference.

mirrorcoder:
Looks good. +1

devlab/tests/config.py
ntymtsiv: We use created images to spawn vms. For testing migration we need more than 1 image.

roman-verchikov:
looks ok, can be merged

PR #394, opened 2015-09-11
Fixed SAIC-433: delete already created flavor on DST if it's needed for instances

realsystem:


Settis:
+1

cloudferrylib/os/actions/check_needed_compute_resources.py
```diff
@@ -61,21 +61,26 @@ def check_in_use_flavor(self, objs, info):
...
-        dst_flavors = [flavor.id for flavor in dst_compute.get_flavor_list()]
+        dst_flavors = dst_compute.get_flavor_list()
+        dst_flavor_ids = [flavor.id for flavor in dst_flavors]
 
         for instance in objs.values():
-            if instance['instance']['flavor_id'] not in src_flavors \
-                    and instance['instance']['flavor_id'] not in dst_flavors:
+            inst_flavor_id = instance['instance']['flavor_id']
+            if inst_flavor_id not in src_flavor_ids \
+                    and inst_flavor_id not in dst_flavor_ids:
```
mirrorcoder: I think that:
flavor_ids = src_flavor_ids+dst_flavor_ids
and 
if inst_flavor_id not in flavor_ids:

miarmak:
Looks good, except of commit message. 
Please, make the first line limited to 50 characters, insert a single blank line after it and also add more detailed commit message below (72 characters max).

cloudferrylib/os/actions/check_needed_compute_resources.py
```diff
@@ -61,21 +61,26 @@ def check_in_use_flavor(self, objs, info):
...
-            if instance['instance']['flavor_id'] not in src_flavors \
-                    and instance['instance']['flavor_id'] not in dst_flavors:
+            inst_flavor_id = instance['instance']['flavor_id']
+            if inst_flavor_id not in flavor_ids:
                 _instance = instance['instance']
                 instance_id = _instance['id']
                 flav_details = \
                     info['instances'][instance_id]['instance']['flav_details']
+                for flavor in dst_flavors:
+                    if flavor.name == flav_details['name']:
```
roman-verchikov: This only verifies whether name exists on destination, even though flavor with this name may be exactly the same. Please consider rewriting by modifying logic in `create_flavor_if_not_exists` method: https://github.com/MirantisWorkloadMobility/CloudFerry/blob/devel/cloudferrylib/os/compute/nova_compute.py#L480

miarmak: Yeah, but we have already checked IDs in line 72... As far as I understand, in this case we need to recreate flavor anyway to have it exactly the same as on the source.

realsystem: Current approach is ok, we don't need to compare other parts except name. If we keep untouched flavor on DST(if it's 100% the same like on SRC) we will see error below:
BadRequest: Invalid flavorRef provided. (HTTP 400)

realsystem: And it's not possible to move logic to nova_compute because it's on early step and we don't have instance info. We need instance info to compare DST flavor and instance flavor details. On SRC flavor is deleted. So, there is only one way - on review.

roman-verchikov:
+1, can be merged.

PR #398, opened 2015-09-14
Fixed SAIC-437: added new pre-migration checks

realsystem:


roman-verchikov:
This one was mistakenly merged into `master` instead of `devel`. Must be merged without changes to `devel` in order to have less pain in merging two branches.

PR #401, opened 2015-09-15
Fixed floating ip network in grizzly and icehouse

ntymtsiv:
Added physical eth1 interface to the floating bridge. This allow access
to the vm's inside of OpenStack directly from the host-server.


mirrorcoder:
+1

roman-verchikov:
+1

PR #406, opened 2015-09-16
[SAIC-456] Add DB config options for identity (keystone) service in CF

mirrorcoder:


realsystem:
+1

cfglib.py
```diff
@@ -174,6 +174,16 @@
...
+    cfg.StrOpt('db_user', default=None,
+               help='user for mysql'),
+    cfg.StrOpt('db_password', default=None,
+               help='password for mysql'),
+    cfg.StrOpt('db_host', default=None,
+               help='host of mysql'),
+    cfg.IntOpt('db_port', default=None,
+               help='port for mysql connection'),
+    cfg.StrOpt('db_connection', default=None,
+               help='driver for connection'),
```
roman-verchikov: I might be missing something, but do we really need DB creds for rabbit?.. My understanding is that for rabbit we only need rabbit-specific creds, not mysql.

miarmak: No, we do not need DB creds for RabbitMQ. For Rabbit we only need Rabbit-specific creds, not mysql. And also we use it only in our checks. So i guess we can remove it form this patch.

cfglib.py
```diff
@@ -273,6 +283,16 @@
                             title='Config service for identity')
 
 src_identity_opts = [
+    cfg.StrOpt('db_user', default=None,
```
miarmak: I guess we can provide some default values here. F.e. 
```
db_user = keystone
db_password = keystone
db_name = keystone
db_connection = mysql+mysqlconnector
db_port = 3306
```

cfglib.py
```diff
@@ -273,6 +283,16 @@
...
+               help='password for mysql'),
+    cfg.StrOpt('db_host', default=None,
+               help='host of mysql'),
+    cfg.IntOpt('db_port', default=None,
+               help='port for mysql connection'),
+    cfg.StrOpt('db_connection', default=None,
+               help='driver for connection'),
     cfg.StrOpt('service', default='keystone',
                help='name service for keystone'),
     cfg.StrOpt('database_name', default='',
```
miarmak: It seems as this option is obsolete and should be renamed to `db_name`.

cfglib.py
```diff
@@ -305,6 +325,16 @@
                               title='Config service for object storage')
 
 src_objstorage_opts = [
```
miarmak: In commit message you've specified only about identity (keystone) DB config options adding. Remove it from this patch OR add info in the commit message about it please.

cfglib.py
```diff
@@ -305,6 +325,16 @@
...
+               help='password for mysql'),
+    cfg.StrOpt('db_host', default=None,
+               help='host of mysql'),
+    cfg.IntOpt('db_port', default=None,
+               help='port for mysql connection'),
+    cfg.StrOpt('db_connection', default=None,
+               help='driver for connection'),
     cfg.StrOpt('service', default='swift',
                help='service name for object storage'),
     cfg.StrOpt('database_name', default='',
```
miarmak: The same issue as above about `db_name`.

cfglib.py
```diff
@@ -331,6 +361,16 @@
                           title='Config RabbitMQ for source cloud')
 
 dst_rabbit_opts = [
```
miarmak: And everything as above for DST.

cfglib.py
mirrorcoder: Default params must be None and cloudferry will use _mysql sections creds

miarmak: OK

cfglib.py
roman-verchikov: @mirrorcoder please remove config options

roman-verchikov:
+1, can be merged.

PR #413, opened 2015-09-18
Fixed SAIC-51 partially: pep8

realsystem:


mirrorcoder:
+1

miarmak:
Please rebase

miarmak:
Seems looks not bad.
 
But as for me (and PEP8 recommendations in https://www.python.org/dev/peps/pep-0008/#maximum-line-length), the preferred way of wrapping long lines is by using Python's implied line continuation inside parentheses, brackets and braces. Long lines can be broken over multiple lines by wrapping expressions in parentheses. These should be used in preference to using a backslash for line continuation. But anyway, it is not a strict rule :wink:

Unit and functional tests have been passed and seems everything is OK. 
So, +1 from me.

roman-verchikov:
Looks good, I think it's a good time to merge

PR #415, opened 2015-09-18
[SAIC-463] Fix CheckCloud action

miarmak:
Add test private network creation with subnet and use it to spawn test
VM. It is needed for cases, when there are multiple networks exist on
the cloud, that CF gonna check. And also to check booting VM in the
particular private network.


roman-verchikov:
Looks good, let's wait for functional tests results

roman-verchikov:
can be merged.

PR #416, opened 2015-09-19
SAIC-462 Regions support

roman-verchikov:
Adds regions functionality support to CF.


roman-verchikov:
Verified on staging, can be merged.

PR #419, opened 2015-09-21
SAIC-468 Config option for DB backend host

roman-verchikov:
Allows user to specify particular MySQL backend host when MySQL is deployed
on multiple nodes, and multiple MySQL nodes are hidden behind a VIP. In this
scenario providing VIP in `config.dst_mysql.db_host` will break mysqldump
which requires to be run locally on particular DB host. Patch allows to specify
`config.migrate.mysqldump_host` which workarounds this problem.


roman-verchikov:
Verified on staging, can be merged.

PR #420, opened 2015-09-22
Saic 469 release fix

roman-verchikov:


roman-verchikov:
Invalid. Should be merged into devel

PR #432, opened 2015-09-23
Update config.ini with identity migration options

roman-verchikov:
Adds description for `optimize_user_role_fetch` config option.

### DIFF
--- a/configs/config.ini
+++ b/configs/config.ini
@@ -158,6 +158,11 @@ path_scenario = scenario/migrate.yaml
 # rollback.
 mysqldump_host = <specific MySQL DB hostname or IP address, not VIP>
 
+# Uses low-level DB requests if set to True, may be incompatible with more
+# recent versions of Keystone. Tested on grizzly, icehouse and juno. True by
+# default
+optimize_user_role_fetch = [True|False]
+
 
 #==============================================================================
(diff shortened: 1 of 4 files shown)


configs/config.ini
```diff
@@ -158,6 +158,11 @@ path_scenario = scenario/migrate.yaml
 # rollback.
 mysqldump_host = <specific MySQL DB hostname or IP address, not VIP>
 
+# Uses low-level DB requests if set to True, may be incompatible with more
+# recent versions of Keystone. Tested on grizzly, icehouse and juno. True by
+# default
+optimize_user_role_fetch = [True|False]
+
```
miarmak: We still have one more config template in `devlab`... Maybe we need to add this option there too?

miarmak:
Looks good overall. Only 1 question inline.

configs/config.ini
Settis: Agree. It can speedup our CI.

Settis:
+1

configs/config.ini
roman-verchikov: `optimize_user_role_fetch` is enabled by default, so it's already there.

roman-verchikov:
can be merged

PR #433, opened 2015-09-23
SAIC-482 : During CF functional testing, assertion required during no expected r…

raiesmh08:
During CF functional testing - when I run test "test_migrate_keystone_user_tenant_roles" without initiation migration. This test is Pass.
As migration is not initiated, and no expected user of source is available on destination.
In this case this test should Fail, raising some message like -
"Either migration is not initiated or was not successful for user resource."
So an assertion is required during no expected resource is available on Destination.


roman-verchikov:
Please squash commits into one

roman-verchikov:
Looks good, can be merged.

PR #435, opened 2015-09-24
SAIC-398 Added check router ips are not overlapping

ntymtsiv:
Check gw ips of external router on src and dst are different.

### DIFF
--- a/devlab/tests/functional_test.py
+++ b/devlab/tests/functional_test.py
@@ -14,9 +14,19 @@
 
 import config
 import logging
+import sys
+import os
 import unittest
 from generate_load import Prerequisites
 from filtering_utils import FilteringUtils
+
+
+def get_cf_root_folder():
+    return os.path.dirname(os.path.dirname(os.path.split(__file__)[0]))
+
+sys.path.append(get_cf_root_folder())
+import cfglib
+cfglib.init_config(os.path.join(get_cf_root_folder(), config.cloud_ferry_conf))
 
 
 def suppress_dependency_logging():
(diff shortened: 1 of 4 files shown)


devlab/tests/generate_load.py
```diff
@@ -824,14 +830,6 @@ def wait_until_vms_all_deleted():
             except Exception as e:
                 print "Tenant %s failed to delete: %s" % (tenant['name'],
                                                           repr(e))
-        sgs = self.neutronclient.list_security_groups()['security_groups']
```
Settis: Why are you leave security groups while do `clean_objects()`?

ntymtsiv: This piece of code is duplicated. We still have cleaning for security_groups.

Settis: Got it.

Settis:
+1

devlab/tests/functional_test.py
```diff
@@ -14,11 +14,21 @@
...
 from generate_load import Prerequisites
 from filtering_utils import FilteringUtils
 
 
+def get_cf_root_folder():
+    return os.path.dirname(os.path.dirname(os.path.split(__file__)[0]))
+
+sys.path.append(get_cf_root_folder())
+import cfglib
+cfglib.init_config(os.path.join(get_cf_root_folder(), config.cloud_ferry_conf))
```
roman-verchikov: Let's better pass configuration.ini as argument to the tests if possible. Relying on config's position from the code is very optimistic.

ntymtsiv:
Path to CloudFerry config is configurable, it specified in config.py.
Do you mean pass config like this: "nosetests --tc-file configuration.ini"?
It will affect all scripts for CI.

roman-verchikov:
Ok, I see. We need to refactor this separately. +1 for now

PR #437, opened 2015-09-24
[SAIC-486] change_router_ips=False crush migration

Settis:
Case when change_router_ips=False and clean_router_ips_stub=True.


roman-verchikov:
Looks good, next time don't forget to run pylint before submitting. Can be merged.

PR #452, opened 2015-10-01
SIAC-521 Existing role migration

roman-verchikov:
Fixes error during migration when the same role already exists on
destination for any reason.

### DIFF
--- a/cloudferrylib/os/identity/keystone.py
+++ b/cloudferrylib/os/identity/keystone.py
@@ -17,6 +17,7 @@ import pika
 import ast
 
 import keystoneclient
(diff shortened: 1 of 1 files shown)


cloudferrylib/os/identity/keystone.py
```diff
@@ -624,9 +625,14 @@ def _upload_user_tenant_roles(self, user_tenants_roles, users, tenants):
                     role = _role['role']
                     if role['name'] in exists_roles:
```
miarmak: In this case what is the purpose of the `exists_roles` list? I thought this issue should be resolved here...

roman-verchikov:
can be merged

PR #457, opened 2015-10-02
[SAIC-308]Migrate quotas network

mirrorcoder:
It has a function of getting quotas and their loading.
Implemented unit tests.

### DIFF
--- a/cfglib.py
+++ b/cfglib.py
@@ -334,7 +334,11 @@ src_network_opts = [
     cfg.StrOpt('db_connection', default=None,
                help='Neutron DB connection type'),
     cfg.StrOpt('db_user', default=None,
-               help="DB user for the networking backend")
+               help="DB user for the networking backend"),
+    cfg.StrOpt('get_all_quota', default=False,
+               help="If False, then qoutas will be taken only customs,"
+                    " else All quotas including default and custom")
+
 ]
 
 src_objstorage = cfg.OptGroup(name='src_objstorage',
@@ -510,7 +514,10 @@ dst_network_opts = [
     cfg.StrOpt('db_connection', default=None,
                help='Neutron DB connection type'),
     cfg.StrOpt('db_user', default=None,
-               help="DB user for the networking backend")
+               help="DB user for the networking backend"),
+    cfg.StrOpt('get_all_quota', default=False,
+               help="If False, then qoutas will be taken only customs, "
+                    "else All quotas including default and custom")
 ]
 
 dst_objstorage = cfg.OptGroup(name='dst_objstorage',
--- a/cloudferrylib/os/actions/check_config_quota_neutron.py
+++ b/cloudferrylib/os/actions/check_config_quota_neutron.py
@@ -1,0 +1,71 @@
+# Copyright (c) 2015 Mirantis Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the License);
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an AS IS BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or
+# implied.
+# See the License for the specific language governing permissions and#
+# limitations under the License.
+from cloudferrylib.base.action import action
+from cloudferrylib.utils import utils as utl
+
+
+LOG = utl.get_log(__name__)
+
+
+class CheckConfigQuotaNeutron(action.Action):
+    """
+    Checking config quotas between src and dst clouds.
+    If all tenants have customs quotas then
+    different configurations does not matter.
+    """
+    def run(self, info=None, **kwargs):
+        src_cloud = self.src_cloud
+        dst_cloud = self.dst_cloud
+        network_src = src_cloud.resources[utl.NETWORK_RESOURCE]
+        identity_dst = src_cloud.resources[utl.IDENTITY_RESOURCE]
+        network_dst = dst_cloud.resources[utl.NETWORK_RESOURCE]
+        tenants_src = self.get_src_tenants()
+        list_quotas = network_src.list_quotas()
+        tenants_without_quotas = self.get_tenants_without_quotas(tenants_src,
+                                                                 list_quotas)
+        if not tenants_without_quotas:
+            LOG.info("On SRC cloud all tenants "
+                     "have custom quotas for network")
+            LOG.info("Difference between clouds configs "
+                     "default quotas will not calculated")
+            LOG.info("Migration can proceed")
+            return
+        LOG.info("Checking default quota "
+                 "configuration on source and destination cloud")
+        quot = network_src.show_quota(tenants_without_quotas[0])
+        dst_temp_tenant = identity_dst.create_tenant("Test Tenant For Quotas")
+        quot_default_dst = network_dst.show_quota(dst_temp_tenant.id)
+        is_configs_different = False
+        identity_dst.delete_tenant(dst_temp_tenant)
+        for item_quot, val_quot in quot.iteritems():
+            if val_quot != quot_default_dst[item_quot]:
+                is_configs_different = True
+                LOG.info("Item %s in quotas is different "
+                         "(SRC CLOUD: %s, DST CLOUD: %s)" %
+                         (item_quot,
+                          val_quot,
+                          quot_default_dst[item_quot]))
+        if not is_configs_different:
+            LOG.info("Configs on clouds is equals")
+
+    def get_tenants_without_quotas(self, tenants_src, list_quotas):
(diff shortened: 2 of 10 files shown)


cloudferrylib/os/actions/check_config_quota_neutron.py
```diff
@@ -0,0 +1,64 @@
...
+# See the License for the specific language governing permissions and#
+# limitations under the License.
+from cloudferrylib.base.action import action
+from cloudferrylib.utils import utils as utl
+
+
+LOG = utl.get_log(__name__)
+
+
+class CheckConfigQuotaNeutron(action.Action):
```
roman-verchikov: Please add docstring, should include:
 - Overview of the task (what it does)
 - Task dependencies (does it depend on any other task)
 - Task dependents (do other tasks depend on this one)
 - Config options required used

cloudferrylib/os/network/neutron.py
```diff
@@ -106,6 +107,44 @@ def read_info(self, **kwargs):
...
+
+    def upload_quota(self, quota):
+        identity = self.identity_client
+        for q_name, q_val in quota.iteritems():
+            tenant_id = identity.get_tenant_id_by_name(q_name)
+            self.neutron_client.update_quota(tenant_id, q_val)
+
+    def create_quota(self, tenant_id, quota):
+        return self.neutron_client.update_quota(tenant_id, quota)
+
```
roman-verchikov: I would recommend moving this functionality into separate class, clearly quotas are an independent entity

cloudferrylib/os/network/neutron.py
```diff
@@ -106,6 +107,44 @@ def read_info(self, **kwargs):
...
+        }
+        if self.config.network.get_all_quota:
+            for t_id, t_val in tenants.iteritems():
+                data[t_val] = self.neutron_client.show_quota(t_id)
+        else:
+            for t in self.neutron_client.list_quotas()['quotas']:
+                if (not tenant_id) or (tenant_id == t['tenant_id']):
+                    tenant_name = self.identity_client.\
+                        try_get_tenant_name_by_id(t['tenant_id'])
+                    del t['tenant_id']
```
roman-verchikov: `del` is a sign of poor design, please refactor. It seems `del` can be omitted by creating a separate object with all required information in `data[tenant_name]`.

cloudferrylib/os/actions/check_config_quota_neutron.py
```diff
@@ -0,0 +1,64 @@
...
+        list_quotas = network_src.list_quotas()
+        not_custom_quotas = self.get_tenants_without_quotas(tenants_src,
+                                                            list_quotas)
+        if not not_custom_quotas:
+            LOG.warning("On SRC cloud all tenants "
+                        "have custom quotas for network")
+            LOG.warning("Difference between clouds configs "
+                        "default quotas will not calculated")
+            return
+        LOG.warning("Check configs on clouds")
```
roman-verchikov: * Change to `LOG.info`, since contents are definitely not a warning
* Update message to something meaningful, maybe something like "Checking default quota configuration on source and destination cloud"

cloudferrylib/os/actions/check_config_quota_neutron.py
```diff
@@ -0,0 +1,64 @@
...
+        src_cloud = self.src_cloud
+        dst_cloud = self.dst_cloud
+        network_src = src_cloud.resources[utl.NETWORK_RESOURCE]
+        identity_dst = src_cloud.resources[utl.IDENTITY_RESOURCE]
+        network_dst = dst_cloud.resources[utl.NETWORK_RESOURCE]
+        tenants_src = self.get_src_tenants()
+        list_quotas = network_src.list_quotas()
+        not_custom_quotas = self.get_tenants_without_quotas(tenants_src,
+                                                            list_quotas)
+        if not not_custom_quotas:
```
roman-verchikov: Double negation (not not) is not readable. Please refactor

cloudferrylib/os/actions/check_config_quota_neutron.py
```diff
@@ -0,0 +1,64 @@
...
+            LOG.warning("On SRC cloud all tenants "
+                        "have custom quotas for network")
+            LOG.warning("Difference between clouds configs "
+                        "default quotas will not calculated")
```
roman-verchikov: * The message seem to be more informational rather than warning, please change to `LOG.info`
* Message contents are not clear. Does it require user attention? Can migration proceed? Please update message to be more descriptive.

cfglib.py
```diff
@@ -334,7 +334,10 @@
     cfg.StrOpt('db_connection', default=None,
                help='Neutron DB connection type'),
     cfg.StrOpt('db_user', default=None,
-               help="DB user for the networking backend")
+               help="DB user for the networking backend"),
+    cfg.StrOpt('get_all_quota', default=False,
+               help="If False, then qoutas will be taken only custom")
```
roman-verchikov: Option name and description are not clear. Please change.

cfglib.py
```diff
@@ -510,7 +513,9 @@
...
```
roman-verchikov: Option name and description are not clear. Please change.

roman-verchikov:
* Please fix inlines;
* Please update commit message to be more descriptive;

roman-verchikov:
@ntymtsiv please verify

cfglib.py
ntymtsiv: Please add new config parameters in configs/config.ini with description.

scenario/cold_migrate.yaml
```diff
@@ -6,6 +6,7 @@ preparation:
   - pre_migration_test:
       - act_get_filter: True
       - act_check_filter: True
+      - check_config_quota_neutron: True
```
ntymtsiv: Add this check in devlab/tests/scenarios/cold_migration.yaml as well

ntymtsiv:
Works good. Custom quotas migrate successfully. If default quotas on dst and src are different, we see warning message.

cloudferrylib/os/network/neutron.py
mirrorcoder: It is not necessary in my opinion. Do I need to make in separate entities if the network subnet, router, and etc?

cloudferrylib/os/actions/check_config_quota_neutron.py
```diff
@@ -0,0 +1,71 @@
...
+        for item_quot, val_quot in quot.iteritems():
+            if val_quot != quot_default_dst[item_quot]:
+                is_configs_different = True
+                LOG.info("Item %s in quotas is different "
+                         "(SRC CLOUD: %s, DST CLOUD: %s)" %
+                         (item_quot,
+                          val_quot,
+                          quot_default_dst[item_quot]))
+        if not is_configs_different:
+            LOG.info("Configs on clouds is equals")
```
roman-verchikov: With current implementation migration will proceed regardless of results of the task, but it should stop if default quotas on source and destination are different.

cloudferrylib/os/network/neutron.py
roman-verchikov: > Do I need to make in separate entities if the network subnet, router, and etc?

Yes at some point, but not with this pull request. Separation of concerns is one of the primary object oriented design patterns, and [blob objects](https://sourcemaking.com/antipatterns/the-blob) are one of the most common anti-patterns. By introducing more functionality (adding quota functionality) into a blob class (`NeutronNetwork` in this case) you simply follow traditional anti-pattern instead of resolving it.

cloudferrylib/os/actions/check_config_quota_neutron.py
mirrorcoder: Really? Can put depending on the option in the config?

cloudferrylib/os/network/neutron.py
miarmak: I agree with @roman-verchikov here, cause our `NeutronNetwork` class stretched out on the extremely huge amount of LOC (~1300). And also, in the `cloudferrylib/os/network/neutron.py` module we've already started doing such separation: https://github.com/MirantisWorkloadMobility/CloudFerry/blob/devel/cloudferrylib/os/network/neutron.py#L1327

cloudferrylib/os/actions/check_config_quota_neutron.py
miarmak: @roman-verchikov As i remember, we decided to:
- get default quota with our manipulations
- print warning about difference
- proceed with migration using default quotas from source (if there is no custom quota)
Isn't it?

miarmak:
I would also suggest updating commit message with more detailed and meaningful description.

cloudferrylib/os/actions/check_config_quota_neutron.py
roman-verchikov: Ok, let's leave warning, it's going to be too complicated to verify if all objects in source don't fit into quotas in destination cloud.

roman-verchikov:
can be merged

PR #459, opened 2015-10-02
Fix for SAIC-186.

raiesmh08:
Currently log configuration file is hardcoded in
'CloudFerry/cloudferrylib/utils/utils.py'.
Instead log config should be available for configuration from the
general CloudFerry configuration file.

Implementation of will go in 2 part -

1. Logging customization
2. Hard coding removal

For the time this part implements point 1.


raiesmh08:
In progress

cloudferrylib/utils/utils.py
```diff
@@ -32,6 +32,10 @@
 import yaml
 from logging import config
 
+import cfglib
+
+# Initialize default config
+cfglib.initialize()
```
roman-verchikov: What if `utils.py` never gets imported? I would suggest that you move all the logging-related logic into a separate module/class.

roman-verchikov:
Please provide description for the new option in `configs/config.ini`

roman-verchikov:
Please update commit message to follow pattern:
```
<One-line title describing the feature/bugfix>

<multiline description of what was done>
```

raiesmh08:
Leaving it for the time. If time permits I will be looking it soon :)

PR #461, opened 2015-10-05
SAIC-515 Added test for migration with two ip pools

ntymtsiv:
Refactored 'create_networks' method in generate load. Now 'subnet'
is not separate parameter in config, it attribute of 'networks'.
This change allows create few subnets for network.
    Removed hardcode name of external network and subnet. Attribute of
network 'real_network' is used for specifying real network. This fix
allows create more than one external netoworks.


plomakin:

PR #468, opened 2015-10-06
[SAIC-531] Fix networks migration

miarmak:
Fix issue with Neutron networks migration for cases, when network has
attribute `provider:physical_network`.


roman-verchikov:
+1, can be merged

PR #470, opened 2015-10-07
SAIC-435,436,503,504 Non-filtered image migration

roman-verchikov:
When migration is filtered by tenant, only filtered tenant migrated
originally.  Glance images which belong to tenants which are not
specified in filter were not migrated and this caused glance image
migration. This patch resolves this issue and refactors filter logic
separating it from migration.


roman-verchikov:
Verified in staging, all good

roman-verchikov:
can be merged

PR #471, opened 2015-10-07
[SAIC-477] Add floating IPs check

miarmak:
Add floating IPs checking to the `check_networks` action. Patch provides
handling duplicates in floating IPs. Check if floating IP overlaps with
the destination cloud. If floating IP is exactly the same as on the
source cloud (checking by address, tenant, network) check passes. If
difference have been found - exception  will be raised.

Also check if floating IP is not busy (i.e. is not associated to VM on
SRC and DST at the same time) on both environments.

### DIFF
--- a/cloudferrylib/os/actions/check_networks.py
+++ b/cloudferrylib/os/actions/check_networks.py
@@ -27,7 +27,7 @@ LOG = utils.get_log(__name__)
 
 class CheckNetworks(action.Action):
     """
-    Check networks segmentation ID and subnets overlapping.
+    Check networks segmentation ID, subnets and floating IPs overlapping.
 
     Raise exception (AbortMigrationError) if overlapping has been found.
     It must be done before actual migration in the 'preparation' section.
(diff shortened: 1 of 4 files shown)


cloudferrylib/os/actions/check_networks.py
```diff
@@ -166,3 +183,43 @@ def check_segmentation_id_overlapping(self, dst_seg_ids):
...
+        VM on SRC and DST at the same time) on both environments.
+
+        :raise AbortMigrationError: If FloatingIp overlaps with the DST.
+        """
+
+        # Check if it is the same Floating IP
+        floatings_are_the_same = (
+            self.tenant == dst_floating_ip.tenant and
+            self.network == dst_floating_ip.network and
+            self.net_tenant == dst_floating_ip.net_tenant)
```
roman-verchikov: I'd suggest that you implement `__eq__` method and then use simple `self == other` to compare FIPs for equity

miarmak: Nice idea, thanks!

miarmak:
Updated, rebased, tested and verified on devlab and staging.

roman-verchikov:
Looks good to me, +1

roman-verchikov:
Can be merged.

PR #472, opened 2015-10-07
SAIC-525 Reafactored cleanup in generate load

ntymtsiv:
- Splited cleanup method in saprate methods
   - Moved cleanup methods into separate class
   - Added hadling for resource creating which
     already exist (run cleanup method for current
     resource and create one more time)
   - Changed Exception, which raised when resource
     not found. Now instead of IndexError is raised
     NotFound Error
   - Refactored cleanup methods. Now try to delete
     resource only when:
     1) Resource name specified in the config file
     2) Resource with given name exists on the cloud

### DIFF
--- a/devlab/tests/functional_test.py
+++ b/devlab/tests/functional_test.py
@@ -17,7 +17,7 @@ import logging
 import sys
 import os
 import unittest
-from generate_load import Prerequisites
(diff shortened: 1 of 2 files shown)


plomakin:

PR #474, opened 2015-10-08
Upgrade Glance Client to 1.1.0

mirrorcoder:
Fix get_ref_image()
Upgrade glance client


requirements.txt
```diff
@@ -20,5 +19,6 @@ pyyaml
 redis
 requests==2.2.1
 sqlalchemy
+oslo.utils==1.9.0
```
miarmak: Why do we need this package?

cloudferrylib/os/image/glance_image.py
```diff
@@ -144,7 +144,10 @@ def get_image_status(self, image_id):
 
     def get_ref_image(self, image_id):
         try:
-            return self.glance_client.images.data(image_id)._resp
+            return getattr(self.glance_client.images.data(image_id),
+                           '_resp',
```
miarmak: no way to get rid of it?

cloudferrylib/utils/iterator_as_file.py
```diff
@@ -0,0 +1,30 @@
+# Copyright (c) 2014 Mirantis Inc.
```
miarmak: 2015 :wink:

cloudferrylib/utils/iterator_as_file.py
```diff
@@ -0,0 +1,30 @@
...
+# distributed under the License is distributed on an AS IS BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or
+# implied.
+# See the License for the specific language governing permissions and#
+# limitations under the License.
+
+
+from utils import get_log
+
+LOG = get_log(__name__)
```
miarmak: We don't use this variable

requirements.txt
roman-verchikov: I assume it was simply moved from above ;) This change is not needed though, so I would recommend reverting it

ntymtsiv:
-1, mirrorcoder  wants to merge 1 commit into MirantisWorkloadMobility:master
Create pull request to devel branch

PR #478, opened 2015-10-10
SAIC-557 Glance user in metadata

roman-verchikov:
Glance images may have user ID stored in metadata, but user may not be
present on destination cloud. In order to avoid migration error
additional pre-migration check was added which verifies all users from
source are available on destination. This case is only applicable for
LDAP-backed clouds, since users usually are not created from keystone.


roman-verchikov:
Verified in staging, can be merged

PR #480, opened 2015-10-12
SAIC-464 Added test for vm booted from volume

ntymtsiv:
Added test for case. when vm booted from volume. This test
enabled only for case, when source env is icehouse or higher,
because early versions of cinder does not support creating
volume from image (in case nfs backend).


plomakin:

PR #293, opened 2015-07-25
Fixed test for keypairs

ntymtsiv:

PR #481, opened 2015-10-12
SAIC-559 OSX compatibility for generate_config.sh

roman-verchikov:
Added compatibility for BSD sed so that generate_config.sh can be used
on OSX.


roman-verchikov:
overlaps with #293

roman-verchikov:
Tested in ubuntu and osx, GNU bash 4.2.25(1) and GNU bash 3.2.53(1), can be merged.

PR #484, opened 2015-10-13
upgrade of python-cinderclient version

toha10:
Python-cinderclient has updated from 1.0.9 to 1.1.0
version to have an ability to change bootable flag of volume.
__patch_option_bootable_of_volume, which uses mysqlconnector
replaced by api call

### DIFF
--- a/cloudferrylib/os/storage/cinder_storage.py
+++ b/cloudferrylib/os/storage/cinder_storage.py
@@ -14,6 +14,7 @@
 
 
 from cinderclient.v1 import client as cinder_client
+from cinderclient import exceptions as cinder_exc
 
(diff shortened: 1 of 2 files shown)


toha10:
Looks like this api functionality ('os-set_bootable' action) appeared only in Juno release.
It's not covering all migration cases, f.e. Grizzly - IceHouse and because of that we need to decide:
do we discard this pull request or will we add some kind of option (or maybe automatic version checking) that will help us to choose right method: api method 'set_bootable' or use DB ('__patch_option_bootable_of_volume') ?

Settis:
Nice! +1

roman-verchikov:
Looks good, can be merged.

PR #487, opened 2015-10-14
Modify setup.py to take version from env

mgrygoriev:
This helps easily upload CloudFerry code to PyPi after release is made.


setup.py
```diff
@@ -14,12 +14,14 @@
...
 
+import os
+
 from setuptools import setup
 from setuptools import find_packages
 from pip.req import parse_requirements
 
 setup(name='CloudFerry',
-      version=open('version.txt').read().strip(),
+      version=str(os.environ['CLOUDFERRY_VERSION']),
```
roman-verchikov: `os.environ` returns string already, there's no need to convert it explicitly.

roman-verchikov:
+1, can be merged.

PR #488, opened 2015-10-14
SAIC-540 Floating IP allocated in different external networks

roman-verchikov:
Previously network migration logic assumed external networks only belong to admin tenant,
which is not the case for the most production environments. This patch adds support for
floating IPs allocated in external networks which belong to different tenants.

### DIFF
--- a/cloudferrylib/base/resource.py
+++ b/cloudferrylib/base/resource.py
@@ -41,6 +41,11 @@ class Resource(object):
     def restore(self):
         pass
 
+    def required_tenants(self):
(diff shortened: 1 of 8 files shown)


miarmak:
Oh sorry, I've commented in actual commit instead of pull reauest.

miarmak:
And also, I am still a supporter of separation Floating IPs issues and Cached functionality in different pull requests to simplify every pull request and commit history :wink:

roman-verchikov:
can be merged due to a number of blockers it resolves and all functional tests pass

PR #491, opened 2015-10-16
[SAIC-481] Fix Floating IPs filtering

miarmak:
Remove redudant and bugs-forming check in the `upload_floatingips`
method of `NeutronNetwork` resource. There are two reasons for it:
 1. We have wrong tenant id to compare (we try to compare tenant_id
    from source with tenant_id on dest and of course they are different,
    because we do not migrate original tenants IDs);
 2. We already performed this filtering, when obtained information about
    migrated Neutron items (filtering has been already done in the
    `read_info` method, so we do not need to do this twice).


roman-verchikov:
+1

roman-verchikov:
can be merged

PR #496, opened 2015-10-19
Removes unnecessary paramiko debug logs from tests

roman-verchikov:
Drops `paramiko.transport` debug logs when functional tests are run.
These have no value to the user and only complicate funtional test
failures.


roman-verchikov:
Can be merged since it doesn't touch any functionality.

PR #499, opened 2015-10-21
change of keystone client authorization logic

toha10:
When we get keystoneclient instance using only user/pass credentials, during further operations keystone client will use internal endpoint for API requests. This way we need to have an access internal cloud network from CF host and it brake initial conceptions of CF tool 'use only public network'. This fix adds additional authorization of keystone client using token and endpoint,which allows us to use some
specific endpoint during API requests.

### DIFF
--- a/cloudferrylib/os/identity/keystone.py
+++ b/cloudferrylib/os/identity/keystone.py
@@ -106,7 +106,7 @@ class KeystoneIdentity(identity.Identity):
 
     @property
     def keystone_client(self):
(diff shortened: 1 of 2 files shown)


miarmak:
Seems pretty similar to that one we had before) 
Looks good, but I would suggest testing on staging.

plomakin:
+1 from staging lab

roman-verchikov:
+1, can be merged.

PR #506, opened 2015-10-22
Modify setup.py to take version from file

mgrygoriev:
It's needed to upload Cloudferry to PyPI


version.txt
```diff
@@ -0,0 +1 @@
+1.7.0
```
miarmak: Will our Release Builder automatically update this file every release?

setup.py
```diff
@@ -14,14 +14,12 @@
...
 
-import os
-
 from setuptools import setup
 from setuptools import find_packages
 from pip.req import parse_requirements
 
 setup(name='CloudFerry',
-      version=os.environ['CLOUDFERRY_VERSION'],
+      version=open('version.txt').read().strip(),
```
miarmak: In this case file remains open. I guess it would be better to close it and free up any system resources taken up by the open file (https://docs.python.org/2/tutorial/inputoutput.html).

version.txt
mgrygoriev: Yes, it will

roman-verchikov:
can be merged.

PR #510, opened 2015-10-26
Fix for network resource

toha10:
Sometimes we can't get appropriate network, subnet or router after migration by resource hash. This way we can get None value for existing network or subnet. This fix covers this cases in get_new_extnet_id and add_router_interfaces methods.

### DIFF
--- a/cloudferrylib/os/network/neutron.py
+++ b/cloudferrylib/os/network/neutron.py
@@ -1202,17 +1202,25 @@ class NeutronNetwork(network.Network):
         for snet_id in src_router['subnet_ids']:
             snet_hash = self.get_res_hash_by_id(src_snets, snet_id)
             src_net = self.get_res_by_hash(src_snets, snet_hash)
-            ex_snet = self.get_res_by_hash(dst_snets, snet_hash)
+
             if src_net['external']:
                 LOG.debug("NOT connecting subnet '%s' to router '%s' because "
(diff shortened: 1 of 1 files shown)


cloudferrylib/os/network/neutron.py
```diff
@@ -1345,11 +1353,12 @@ def get_resource_hash(neutron_resource, *args):
...
     def get_new_extnet_id(self, src_net_id, src_nets, dst_nets):
         if src_net_id in self.ext_net_map:
-            dst_net_id = self.ext_net_map[src_net_id]
+            return self.ext_net_map[src_net_id]
         else:
             net_hash = self.get_res_hash_by_id(src_nets, src_net_id)
-            dst_net_id = self.get_res_by_hash(dst_nets, net_hash)['id']
-        return dst_net_id
+            dst_net = self.get_res_by_hash(dst_nets, net_hash)
+            if dst_net:
```
roman-verchikov: This only hides the problem, not resolves it. The real problem is that network with the given hash does not exist on destination, and this error must not occur at all.

roman-verchikov:
This seem to be not a resolution but rather an attempt to hide the actual problem. Please rethink.

roman-verchikov:
Closing for now, @toha10 please reopen if or when concerns outlined here get resolved.

PR #512, opened 2015-10-26
[SAIC-587] Hide SSL warnings config param

Settis:


archyufa:
Anton will have the warning fixes going to be removed from cloudferry.py permanently or we will need to add them back once we have "pip install cloudferry" working properly again?
If yes. Can we just keep this fix their in cloudferry.py for future or this may conflict with something ?

roman-verchikov:
Looks ok, but will only work for migration, other tasks (condense, evacuate, etc) will still have SSL warnings. I'd suggest that we have all of those covered with this option. -1 until these are resolved.

Settis:
@archyufa cloudferry.py use fabfile.py function. That means this fix will be available in both running ways:
- fab migrate
- cloudferry migrate

roman-verchikov:
Looks good, +1. Can be merged.

archyufa:
Tested in IAD1. Working well

PR #514, opened 2015-10-27
SAIC-554 cold evacuation support

antonf:
Cold evacuation consists of following steps:
 1) Stop VM if it's running
 2) Detach all ports and volumes
 3) Boot new VM with same ports and volumes
 4) Shutdown new VM
 5) Copy ephemeral disk to new VM
 6) Start new VM
 7) Delete old VM


cloudferrylib/os/compute/cold_evacuate.py
```diff
@@ -0,0 +1,371 @@
...
+    :return: path to ephemeral volume file
+    """
+
+    libvirt_name = getattr(instance, INSTANCE_LIBVIRT_NAME_ATTRIBUTE)
+    instance_block_info = runner.run(
+        "virsh domblklist {}".format(libvirt_name)).split()
+    return utils.get_disk_path(instance, instance_block_info)
+
+
+def cold_evacuate(config, compute_api, network_api, nova_mysql_connector,
```
roman-verchikov: This function has a lot of logic introduced, please cover with unit tests

roman-verchikov:
Looks good, please cover `cold_evacuate` function with unit tests, since it has the most logic introduced.

PR #517, opened 2015-10-27
Adding beginner documentation in markdown format

cburns-mirantis:
Documentation is primarily targeted at deployment engineers starting to work
with cloudferry.


roman-verchikov:
Looks good other than the `nfs` VM usage we discussed previously. Will need to be updated once we have `cloudferry` VM back to vagrant.

PR #523, opened 2015-10-28
SAIC-614 chown temporary directory to current user

antonf:
When creating remote temporary directory with "sudo mktemp" further attempt
to scp file to it will fail if scp is done using non-root user. This patch
adds "sudo chown {user} {tmp_directory}" step to temporary remote dirctory
creation process.


roman-verchikov:
Looks good, +1

roman-verchikov:
Verified in staging, can be merged.

PR #524, opened 2015-10-29
Fixed keypair filtering in func tests

ntymtsiv:
To get all keypairs, which were created via generate_load.py,
method get_users_keypairs logins under each valid user,
which were created by generate_load.py and takes keypairs.list().
For dst clould, users passords updates according to config.py
(because user passwords migrate not correctly)


pavloivanets:
Looks ok
+1

roman-verchikov:
+1, can be merged.

PR #528, opened 2015-10-30
Resolves all flake8 issues, including F401 and F403

roman-verchikov:
Following flake8 errors were ignored previously:
 - F401 - module imported but unused
 - F403 - ‘from module import *’ used; unable to detect undefined names

This patch resolves all these issues.

### DIFF
--- a/cloudferrylib/utils/__init__.py
+++ b/cloudferrylib/utils/__init__.py
@@ -11,7 +11,3 @@
 # implied.
 # See the License for the specific language governing permissions and#
 # limitations under the License.
-
-__author__ = 'mirrorcoder'
-
-from utils import *
--- a/cloud/__init__.py
+++ b/cloud/__init__.py
@@ -1,3 +1,0 @@
-import cloud
-import cloud_ferry
-import os2os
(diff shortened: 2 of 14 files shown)


cloudferrylib/utils/__init__.py
```diff
@@ -11,7 +11,3 @@
 # implied.
 # See the License for the specific language governing permissions and#
 # limitations under the License.
```
miarmak: I guess we can remove this copyright as well.

miarmak:
Looks good overall, except of 1 small comment inline.

cloudferrylib/utils/__init__.py
roman-verchikov: Well this seems to be unrelated to the purpose of the patch, plus it's not 100% clear whether this particular file needs to be copyrighted or not, according to [Apache FAQ](http://www.apache.org/legal/src-headers.html#faq-exceptions):

> Q: What files in an Apache release do not require a license header?
> A: A file without any degree of creativity in either its literal elements or its structure is not protected by copyright law; therefore, such a file does not require a license header. **If in doubt about the extent of the file's creativity, add the license header to the file**.

Since we're in doubt, let's keep the header there :)

roman-verchikov:
Can be merged

PR #531, opened 2015-11-01
Skips down or disabled hosts

roman-verchikov:
- Previoulsy only hosts known to nova as 'down' were skipped, with this
  patch hosts which are 'disabled' are also treated as not available for
  migration;
- Removed config option to skip down or disabled hosts, now these hosts
  are always skipped;
- Resolved pylint warnings and errors in files modified.


roman-verchikov:
Verified in staging, all fine

archyufa:
Verified in staging. IMHO requires extra check added if active VMs spawned on that host.

miarmak:
Looks good, +1.

What about check - it's good idea, but we can implement it as separate pull request...

PR #535, opened 2015-11-02
SAIC-628 Glance image filtering for user outside of filtered tenant

roman-verchikov:
When filtering by tenant is enabled and user specified in migration
config is not member of a filtered tenant, an error occurs when CF tries
to read glance image members belonging to a tenant. This patch provides
resolution to the issue by temporarily adding migration user to filtered
tenant.


roman-verchikov:
Invalid

PR #545, opened 2015-11-05
[SAIC-603] NoneType snet/ext_snet handling

raiesmh08:
From "https://github.com/MirantisWorkloadMobility/CloudFerry/blob/master/
cloudferrylib/os/network/neutron.py#L1201"

we can notice that, during hit of this portion of code, if src_net and ex_snet
is None then it will fail always.

This patch puts fix for it.


MirantisWorkloadMobility:

PR #563, opened 2015-11-10
Returned missed storage host option

toha10:
host option in storage section was renamed to db_host, but this option is used f.e. in case of CEPH backend for volumes for direct connection to storage node for getting data from src cloud and uploading this data to dst storage host using the same option for dst cloud.


MirantisWorkloadMobility:

roman-verchikov:
Looks good, can be merged

PR #620, opened 2015-11-24
[SAIC-453] Extend func tests due to different admin tenants migration

pavloivanets:
Extend functional tests to verify migration
when source and destination migration tenants are different


MirantisWorkloadMobility:

PR #656, opened 2015-12-03
[SAIC-776] Filter routers by tenant

Settis:


Settis:
Wrong implementation. 
I assumed that router can't connect private networks from other tenants, but it is not.

PR #720, opened 2015-12-23
Removes unnecessary output from rsync

roman-verchikov:
Before copying ephemeral storage using `rsync` CF verifies if `rsync` is
installed by calling `rsync --help` which produces a lot of unnecessary
output. This patch reduces amount of output produced by redirecting it
to `/dev/null`.


MirantisWorkloadMobility:

roman-verchikov:
can be merged

PR #793, opened 2016-01-28
Remove file path from logs

roman-verchikov:
Removes file path from logging output in order to save space, and since
that information is often not really valuable.


MirantisWorkloadMobility:

PR #796, opened 2016-01-29
Restart openstack services after VM provision

mgrygoriev:
Check all openstack services after VM provision and restart those which are down


roman-verchikov:
I believe this needs to be done on box level, not from external script, -1 unless you have really good argument for this solution

MirantisWorkloadMobility:

PR #807, opened 2016-02-03
Remove identity from glance with no rollback stage

archyufa:
This patch removes act_identity_trans task from scenario/stages_without_rollback/3_get_and_transfer_images.yaml.

### DIFF
--- a/scenario/stages_without_rollback/3_get_and_transfer_images.yaml
+++ b/scenario/stages_without_rollback/3_get_and_transfer_images.yaml
@@ -8,6 +8,5 @@ preparation:
 process:
   - act_get_filter: True
   - act_check_filter: True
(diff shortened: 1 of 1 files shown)


MirantisWorkloadMobility:

SVilgelm:
I'm not sure that the image's migration will work correctly with this patch.

miarmak:
It works. At least in my devlab) And also I've not seen any dependencies in the code. Except of `2_identity_transporter.yaml` step (scenario) anyway should be done before this one.

And also it seems `migrate_vms.yaml` step also does not need `act_identity_trans` action inside of it anymore. In devlab it works fine without it.

SVilgelm:
:+1: can be merged

PR #808, opened 2016-02-03
[SC-186] Fix copy ephemeral disk when flavor was deleted.

antonf:
If flavor was deleted and then recreated with same ID, nova will
return old flavor for GET request in grizzly clouds. After this patch
code no longer rely on what nova returned in GET request, and use data
from SQL database instead.


MirantisWorkloadMobility:

roman-verchikov:
+1

archyufa:
+1 Tested in staging

PR #838, opened 2016-02-11
[CF-161] VerifyVms shows incorrect results

mirrorcoder:
adding in failed_vms if only vm don't found in dst

### DIFF
--- a/cloudferrylib/os/actions/verify_vms.py
+++ b/cloudferrylib/os/actions/verify_vms.py
@@ -75,7 +75,6 @@ class VerifyVms(action.Action):
                         dst_cmp_inst['flav_details']):
                     LOG.warning("Wrong flav_details of instance %s on DST",
                                 src_inst_id)
-                    failed_vms.append(src_inst_id)
                 if src_inst_info['key_name'] != dst_cmp_inst['key_name']:
                     LOG.warning("Wrong key_name of instance %s on DST",
                                 src_inst_id)
@@ -88,7 +87,6 @@ class VerifyVms(action.Action):
                 if src_inst_info['volumes'] != dst_cmp_inst['volumes']:
                     LOG.warning("Wrong volumes of instance %s on DST",
                                 src_inst_id)
-                    failed_vms.append(src_inst_id)
 
                 # Verify that migrated VM belongs to correct server group
                 if (src_inst_info['server_group'] !=


MirantisWorkloadMobility:

miarmak:
+1

SVilgelm:
:-1: This is not solving the problem, just muting, we have to have an understanding what does correctly migrated instance mean.

miarmak:
I guess **correctly migrated instance** is that instance, that was migrated (recreated) on destination as is. But we have so many cases, when it is not done in this way (recreated flavor or image can differ from source for example by id), but still VM is migrated as was supposed. So IMO in this action that would be fine if we fix it in this way.

cloudferrylib/os/actions/verify_vms.py
```diff
@@ -70,25 +70,20 @@ def run(self, **kwargs):
...
-                    failed_vms.append(src_inst_id)
                 if (src_inst_info['flav_details'] !=
                         dst_cmp_inst['flav_details']):
                     LOG.warning("Wrong flav_details of instance %s on DST",
                                 src_inst_id)
-                    failed_vms.append(src_inst_id)
                 if src_inst_info['key_name'] != dst_cmp_inst['key_name']:
                     LOG.warning("Wrong key_name of instance %s on DST",
                                 src_inst_id)
-                    failed_vms.append(src_inst_id)
```
roman-verchikov: Key name must always stay the same

cloudferrylib/os/actions/verify_vms.py
```diff
@@ -70,25 +70,20 @@ def run(self, **kwargs):
                 if src_inst_info['name'] != dst_cmp_inst['name']:
                     LOG.warning("Wrong name of instance %s on DST",
                                 src_inst_id)
-                    failed_vms.append(src_inst_id)
```
roman-verchikov: VM name must always stay the same

cloudferrylib/os/actions/verify_vms.py
```diff
@@ -70,25 +70,20 @@ def run(self, **kwargs):
...
                 if src_inst_info['key_name'] != dst_cmp_inst['key_name']:
                     LOG.warning("Wrong key_name of instance %s on DST",
                                 src_inst_id)
-                    failed_vms.append(src_inst_id)
                 if (sorted(src_inst_info['interfaces']) !=
                         sorted(dst_cmp_inst['interfaces'])):
                     LOG.warning("Wrong interfaces of instance %s on DST",
                                 src_inst_id)
-                    failed_vms.append(src_inst_id)
```
roman-verchikov: Interfaces must always stay the same, it's one of the most critical points to check and defines migration success more than all other attributes combined.

mirrorcoder:
fix

MirantisWorkloadMobility:

roman-verchikov:
can be merged

PR #844, opened 2016-02-15
[CF-201] Extend functional tests to verify deleted image in dst with the same ID

vladiskuz:
Image will be created on src and dst with the same UUID.
After that deleted from dst before migration with the
generate_load.py script. In this case CF mustcreate new
UUID during migration for this image and migrate it successfully.


MirantisWorkloadMobility:

PR #861, opened 2016-02-24
[CF-214] Exclude non-active images from migration

miarmak:
Sometimes Glance client returns images list, which contains non-active
images in there (f.e. with such statuses as `queued`, `deleted` etc).
But we do not need to retrieve information about them, cause some
actions try to get them from Glance, but fail in some cases (f.e. if
image status is `deleted`). In other cases data of such images will be
just invalid, so we will not migrate them in any case.

This patch excludes such images from the images list and writes it in
the debug log.

### DIFF
--- a/cloudferrylib/os/image/filters.py
+++ b/cloudferrylib/os/image/filters.py
@@ -62,6 +62,10 @@ def public_filter():
     return lambda i: i.is_public
 
 
+def active_filter():
+    return lambda i: i.status == 'active'
+
+
 def tenant_filter(filtered_tenant_id):
     """Filters images not specified in tenant_id section of filters file"""
     return lambda i: (_tenant_filtering_disabled(filtered_tenant_id) or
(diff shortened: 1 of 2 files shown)


MirantisWorkloadMobility:
Consider resolving pylint issues: \
``` \
************* Module tests.cloudferrylib.os.image.test_glance_image
tests/cloudferrylib/os/image/test_glance_image.py:51: [W0212(protected-access), GlanceImageTestCase.setUp] Access to a protected member _resp of a client class

**Comment by MirantisWorkloadMobility** [comment]

File: cloudferrylib/os/image/glance_image.py [hunk 1]
```diff
@@ -200,7 +200,15 @@ def get_image_list(self):
             LOG.info("Filtered images: %s",
                      ", ".join(('%s (%s)' % (i.name, i.id) for i in images)))
 
-        return images
+        result_images = []
```
SVilgelm: Implement is_active filter like is_public in filters.py

miarmak: Done

MirantisWorkloadMobility:

roman-verchikov:
+1, can be merged

PR #887, opened 2016-03-04
Fix for CF-310 : Fixing "MaxAttemptsReached" error

raiesmh08:
When we are going to use retry facility(In terms of decorator) using
'retrying' module, and if within a class (For example A) a method
(For example B) with retry decorator is supposed to be called more than
specified max_attempts, Then for all calls which are coming after
'max_attmpts' number will be failing with a reason "Max attempt reached"

This patch provides fix for above issue.

For more detail of problem, please visit CF-310


MirantisWorkloadMobility:
PEP8 check success, Unittests failed
http://jenkins-cf.vm.mirantis.net:8080/job/cloudferry-unittests-and-style-checks/2791/

roman-verchikov:
that's expected behavior, there's nothing to fix

roman-verchikov:
If you want to retry in endless loop you can set `max_time = UNLIMITED`

PR #889, opened 2016-03-04
[CF-310] Reset retrying during the run function call.

SVilgelm:
The run function of Retry object has to reset the attempts or total time,
otherwise we cannot use the object as a decorator.


MirantisWorkloadMobility:

roman-verchikov:
please add unit tests to verify scenario when function is called multiple times with retry decorator

MirantisWorkloadMobility:

roman-verchikov:
+1, can be merged

PR #896, opened 2016-03-07
CF-319 Remove non-migrated volumes

roman-verchikov:
Steps to reproduce

 1. Set invalid path for bbcp executable in `[bbcp] path` config option
 2. Run volume migration
 3. Notice unhandled exception
 4. Put correct value to `[bbcp] path`
 5. Run volume migration again
 6. Notice volume was skipped as already migrated

Actual behavior

 - Volume is skipped as "migrated" even though it's empty

Expected behavior

 - Volume should be removed on CopyEngine initialization failure


MirantisWorkloadMobility:

PR #915, opened 2016-03-14
[FT] Moved path to scenarion in config

ntymtsiv:
Moved path to scenario in method, which generates
scenario in the test, which verifies work of rollbacks.


MirantisWorkloadMobility:

PR #967, opened 2016-04-12
[CF-378] Add `ext_net_map` check to `check_networks`

miarmak:
Add external networks mapping check to existing checks in the
`CheckNetwork` action. It allows us to have some overlaps, if checking
network's ID is specified in the external network mapping configuration
file (`[migrate] ext_net_map` config option contains path to this file).
In this case network will be matched with another network in the
destination cloud without creating a new one and migration will not fail.


MirantisWorkloadMobility:

SVilgelm:
👍 Can be merged

PR #968, opened 2016-04-13
CF-400 Verify deleted image which have the same IDs as images in SRC

plomakin:
Test scenario:
- Create image in source cloud
- Boot VM from that image
- Create the same (same name, ID, md5, tenant) image in destination cloud
- Delete image from destination cloud using Glance APIs
- Create same image in destination again, but with different ID
- Run VM migration
- Verify that VM was booted from image created in destination cloud in step 5.


MirantisWorkloadMobility:

SVilgelm:
👍 can be merged

PR #974, opened 2016-04-14
CF-358 Need to refactor 'update_filtering_file' method

plomakin:
- Create separate filter file for each tenant in config.py (including admin tenant, and excluding tenants, which are deleted on src ( field "deleted": True in config.py))
- Filter files should have name 'filter_{tenant_name}.yaml'
- Each filter should contain lists of images, instances and volumes, which are belong to given tenant (except instances and images, which should not be included in the filter according to config.py, `images_not_included_in_filter` and `vms_not_in_filter`)
- 'images' section should contain parameter 'dont_include_public_and_members_from_other_tenants': False


MirantisWorkloadMobility:

MirantisWorkloadMobility:

SVilgelm:
👍 Can be merged

PR #1016, opened 2016-05-15
[CF-477] Incorrect validation of config_drive param in functional tests

jaspreetrawel:
Fixed the VM config_drive parameter validation logic. e.g. When enabled, this parameter is set to u'1' on grizzly and u'True' on juno.


jaspreetrawel:
created to master branch by mistake.

PR #1064, opened 2016-06-16
CF-550 Disable API rate limit for grizzly box

roman-verchikov:
All boxes except grizzly have API rate limit disabled. This patch
configures nova on all boxes to disable API rate limit.


MirantisWorkloadMobility:

PR #1098, opened 2016-07-19
[CF-569] Add flavor parameters discovery.

antonf:
This patch add flavor discovery through database. We can't rely on
flavor discovery through API as it hide the ID that is used to link
VM instances to flavors internally. This may lead to nova API
showing that VM with 512Mb of RAM is booted from flavor with ID "1",
but flavor with ID "1" define 64Mb of RAM if someone delete this
flavor after VM is booted and recreate it later with different
parameters.


MirantisWorkloadMobility:

MirantisWorkloadMobility:

PR #1102, opened 2016-07-25
Fix ascii decoding error

SVilgelm:
Decode log messages with `error="replace"` parameter.


MirantisWorkloadMobility:

roman-verchikov:
We should only use UTF-8

PR #1108, opened 2016-07-28
CF-591 Check if bbcp is installed and allowed to be owerwritten

roman-verchikov:
If migration is done with bbcp as copy engine, bbcp is being copied to
both source and destination nodes prior to ephemeral migration, but if
bbcp was previously copied to the node from different user, current
user may not have permissions to overwrite existing bbcp executable and
migration would fail. This patch resolves this issue by reusing existing
bbcp and assigning 777 permissions to the file so that it can be
overwritten by anyone.

### DIFF
--- a/cloudferry/lib/copy_engines/bbcp_copier.py
+++ b/cloudferry/lib/copy_engines/bbcp_copier.py
@@ -28,6 +28,15 @@ CONF = cfg.CONF
 BBCP_PATH = '/tmp/bbcp'
 
 
+def _bbcp_installed(runner, bbcp):
+    try:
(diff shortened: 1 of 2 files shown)


MirantisWorkloadMobility:
PEP8 check failed
http://jenkins-cf.vm.mirantis.net:8080/job/cloudferry-unittests-and-style-checks/3991/

SVilgelm:
👍 Looks good to me

MirantisWorkloadMobility:

roman-verchikov:
can be merged

PR #1132, opened 2016-08-12
[CF-604] Fix setting gateway on networks that don't have FIPs

SVilgelm:
Show a warning instead of stopping the migration if setting a gateway
for the routers fails on external networks that don't have FIPs.


MirantisWorkloadMobility:
All checks successful.

roman-verchikov:
It's not a bug, but a feature, closing as "won't fix"

PR #1133, opened 2016-08-12
CF-602 User quota migration speedup

roman-verchikov:
Improves migration speed for user quotas by removing unnecessary API
requests.


MirantisWorkloadMobility:
All checks successful.

PR #1137, opened 2016-08-24
[CF-618] Transfer files without sudo

SVilgelm:
Modify BaseCopier.runner function to allow using a RemoteRunner without sudo.


MirantisWorkloadMobility:
Some checks failed.
http://jenkins-cf.vm.mirantis.net:8080/job/cloudferry%20checks/66/

SVilgelm:
It's impossible to implements this feature, because we cannot change  the rights of the files on the nfs backend.

PR #1138, opened 2016-08-24
[CF-586] generate load not attaching some volumes to VMs on SRC cloud

jaspreetrawel:
Fix the floating IP configuration for the VMs to be attached to volumes


MirantisWorkloadMobility:
Some checks failed.
http://jenkins-cf.vm.mirantis.net:8080/job/cloudferry%20checks/67/

plomakin:
This changes looks not necessary, we don't validate anything inside these volumes.

PR #1156, opened 2016-09-26
CF-641 Invalid report for tenant migration

roman-verchikov:
Tenant migration resulted in full tenant object being printed in
"Destination ID" field, instead of tenant ID only, which was caused
by incorrect data being passed to notification observer.


MirantisWorkloadMobility:
Some checks failed.
http://jenkins-cf.vm.mirantis.net:8080/job/cloudferry%20checks/166/

roman-verchikov:
can be merged

PR #1162, opened 2016-10-13
CF-648 KeyError during quota migration

roman-verchikov:
Quota migration failed with KeyError traceback because it was trying to
access quotas for non-existing tenant.


MirantisWorkloadMobility:
Some checks failed.
http://jenkins-cf.vm.mirantis.net:8080/job/cloudferry%20checks/201/

PR #1163, opened 2016-10-18
CF-652 Mapping to nothing raises AttributeError

roman-verchikov:
When user specifies mapping in `[migrate] resource_map` file to nothing,
`AttributeError` is risen, for example:

```
   $ cat configs/resource_map.yaml
   # Mapping tenants for migration:
   tenant_map:
     # tenantX: tenantY
```

This patch would raise configuration error in that cases stating that
one cannot assign mapping to nothing.


roman-verchikov:
can be merged

PR #1164, opened 2016-10-19
CF-642 Incomplete migrations of attached volumes report failures

roman-verchikov:
If volume fails to attach to VM for any reason, report would contain
two entries for the same VM:

```
| 2.3 | VM | server2 (<id>) | <dst_id> | success    | None             |
| 2.4 | VM | server2 (<id>) | None     | incomplete | Cannot attach
volume 'f1b16b58-ebc5-4b55-be16-f584978a350c' to VM 'server2': volume
does not exist |
```

The patch resolves the problem by correctly associating VM
migration status in report.


MirantisWorkloadMobility:
Some checks failed.
http://jenkins-cf.vm.mirantis.net:8080/job/cloudferry%20checks/150/

